PR #2, opened 2020-09-08
First commit. with Sign in Page, Auth Guards, Injectors folder structure etc etc and etc

Julias0:


src/app/auth/auth.module.ts
```diff
@@ -0,0 +1,12 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+
+
```
iamtarun1993, line 4: remove this extra line

src/app/auth/sign-in/sign-in.page.ts
```diff
@@ -0,0 +1,126 @@
...
+    await this.loaderService.showLoader();
+
+    const checkEmailExists$ = this.routerAuthService
+      .checkEmailExists(this.fg.controls.email.value)
+      .pipe(
+        catchError(err => {
+          this.handleError(err);
+          return throwError(err);
+        }),
+        shareReplay(),
```
iamtarun1993, line 42: nice...

src/app/core/services/auth.service.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  constructor(
+    private storageService: StorageService,
+    private tokenService: TokenService,
+    private apiService: ApiService,
+    private dataTransformService: DataTransformService
+  ) { }
+
+  getEou(): Promise<ExtendedOrgUser> {
+    return this.storageService.get('user');
```
iamtarun1993, line 23: not sure, but we also nee to pull ExtendedOrgUser from api if user is not present in storageService

Julias0: This will always be present since we set this when the token is set. Same as out web app / mobile app

PR #3, opened 2020-09-15
saml login

iamtarun1993:


capacitor.config.json
```diff
@@ -9,5 +9,8 @@
       "launchShowDuration": 0
     }
   },
```
iamtarun1993, line 11: will revert this file changes later

src/app/auth/sign-in/sign-in.page.ts
```diff
@@ -27,9 +28,53 @@ export class SignInPage implements OnInit {
...
+      const samlNewRefreshToken$ = this.authService.newRefreshToken(data.refresh_token);
+
+      samlNewRefreshToken$.subscribe(() => {
+        this.router.navigate(['/', 'auth', 'switch-org']);
+      });
+    }
+  };
+
+  handleSamlSignIn (res) {
+    var url = res.idp_url + '&RelayState=MOBILE';
```
Julias0, line 54: use let & const instead if var

Julias0: modify tslint file and use ng lint to have these come up in terminal

Julias0 (approved):
PR lgtm - minor comments

PR #15, opened 2020-09-23
Added verify page

Julias0:


src/app/core/services/router-auth.service.ts
```diff
@@ -72,7 +72,7 @@ export class RouterAuthService {
     this.setClusterDomain(data.cluster_domain);
     const resp = await this.fetchAccessToken(data.refresh_token);
     this.newAccessToken(resp.access_token);
-    return resp;
+    return data;
```
iamtarun1993, line 75: why returning data instead of response ???

iamtarun1993 (approved):

src/app/core/services/router-auth.service.ts
Julias0: Ah this is the same as old app - done because we need refresh token in the response for a few pages.

PR #26, opened 2020-10-05
Added Team Trips page

madhavmansuriya40:
Added team trips pages


src/app/core/services/trip-requests.service.ts
```diff
@@ -45,6 +45,39 @@ export class TripRequestsService {
     );
   }
 
+  getTeamTrips(config: Partial<{ offset: number, limit: number, queryParams: any }> = {
+    offset: 0,
+    limit: 10,
+    queryParams: {
```
Julias0, line 51: pass this in from the team trips component - this shouldn't be default

src/app/fyle/Team/team-trips/team-trips.page.scss
```diff
@@ -0,0 +1,105 @@
+$item-border-color: #e0e0e0;
+$background-color: #fff;
+$menu-botton-color: #000;
+$toolbar-border: #ababab6b;
+
+.trip {
+  &--header {
+    background-color: $background-color;
+    height: 80px !important;
```
Julias0, line 9: avoid using !important. Add a todo here if this is difficult to solve

Julias0 (approved):

PR #53, opened 2020-11-05
my edit report

iamtarun1993:


src/app/core/services/report.service.ts
```diff
@@ -251,4 +251,23 @@ export class ReportService {
...
+  }
+
+  addTransactions(reportId, txnIds) {
+    return this.apiService.post('/reports/' + reportId + '/txns', {
+      ids: txnIds
+    });
+  }
+
+  removeTransaction(rptId, txnId, comment?) {
+    var aspy = {
```
Julias0, line 265: aspy?

Julias0 (approved):

src/app/core/services/report.service.ts
iamtarun1993: this variable used old shared js also

PR #78, opened 2020-12-11
Minor fixes

madhavmansuriya40:
- all component file deleted but forgot to remove import entry
- removed extra elements from HTML left while resolving conflict


madhavmansuriya40:
fixed in @Julias0 PR closing this

PR #86, opened 2020-12-16
Minor fixes

madhavmansuriya40:


src/app/fyle/view-team-expense/view-team-expense.page.ts
```diff
@@ -66,6 +72,9 @@ export class ViewTeamExpensePage implements OnInit {
       switchMap(() => {
         return this.transactionService.getEtxn(txId);
       }),
+      tap(res => {
+        this.reportId = res.tx_report_id;
```
Julias0, line 76: This should be done inside subscribe

madhavmansuriya40: we don't subscribe in ts i think we have used |async

Julias0 (approved):

PR #89, opened 2020-12-17
Android fix 1

iamtarun1993:


capacitor.config.json
```diff
@@ -9,5 +9,8 @@
       "launchShowDuration": 0
```
iamtarun1993, line 9: will remove this change when merging

Julias0 (approved):

PR #101, opened 2020-12-21
Final fixes 8

madhavmansuriya40:


src/app/shared/components/fy-select/fy-select-modal/fy-select-modal.component.html
```diff
@@ -2,7 +2,7 @@
   <ion-toolbar mode="md">
     <ion-title>
       <h1 class="selection-modal--title">
-        Select Items
+        {{selectModalHeader || 'Select Items'}}
```
Julias0, line 5: Do the default value inside ts file

PR #108, opened 2020-12-28
Some camera overlay fixes and show loader in google login fix

iamtarun1993:


src/app/auth/sign-in/sign-in.page.ts
```diff
@@ -159,14 +159,18 @@ export class SignInPage implements OnInit {
   }
 
   googleSignIn() {
-    this.googleAuthService.login().then(data => {
+    this.googleAuthService.login().then(async data => {
+      await this.loaderService.showLoader();
```
Julias0, line 163: ok this whole structure is wrong. We shouldn't mix async await promises and rxjs.

Lets move this wholly into rxjs

Julias0 (approved):

PR #113, opened 2020-12-29
fixes

madhavmansuriya40:


src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -510,23 +510,14 @@ export class AddEditExpensePage implements OnInit {
...
 
       const splitExpensePopover = await this.popoverController.create({
         component: SplitExpensePopoverComponent,
         componentProps: {
-          class: popupTypeItemClass,
           areProjectsAvailable,
           areCostCentersAvailable
         },
-        cssClass: 'split-expense-popover'
+        cssClass: 'dialog-popover'
```
iamtarun1993, line 520: why the cssClass name got changed...
IIRC I put extra required scss for this class in global.scss file

src/app/fyle/add-edit-expense/split-expense-popover/split-expense-popover.component.html
```diff
@@ -1,25 +1,27 @@
-<div class="split-expense-popover split-expense-popover--{{class}}">
```
iamtarun1993, line 1: split-expense-popover--{{class}} is required to make the popover height based on no of element available ..

why this got removed?

madhavmansuriya40: I managed to increase decrease height dynamically in CSS itself, this was written very badly in the old app as well

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
madhavmansuriya40: no sir, we don't need extra CSS I managed it in CSS and that fix went in master long ago

iamtarun1993: oh then its ok.

iamtarun1993 (approved):

PR #116, opened 2020-12-29
global font fix, iOS remove black background, and open decimal keyboard

iamtarun1993:


src/app/fyle/my-profile/my-profile.page.ts
```diff
@@ -71,7 +73,9 @@ export class MyProfilePage implements OnInit {
     private toastController: ToastController,
     private orgUserService: OrgUserService,
     private popoverController: PopoverController
-  ) { }
+  ) {
+    this.ROOT_ENDPOINT = environment.ROOT_URL;
```
Julias0, line 77: Should this give the cluster domain wala Url? If yes this wont work. Lets hit the cached API call to fetch cluster domain and get this from there

Julias0 (approved):

PR #117, opened 2021-01-02
Final fixes 17

madhavmansuriya40:


src/app/core/services/trip-request-custom-fields.service.ts
```diff
@@ -29,7 +29,7 @@ export class TripRequestCustomFieldsService {
     }).sort(this.sortCustomFieldsByType);
   }
 
-  @Cacheable()
+  // @Cacheable()
```
Julias0, line 32: ?

Julias0 (approved):

src/app/core/services/trip-request-custom-fields.service.ts
madhavmansuriya40: was testing something will remove this in next PR

PR #150, opened 2021-01-15
resolved TODOs - phase 1

madhavmansuriya40:


src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -1427,18 +1427,28 @@ export class AddEditExpensePage implements OnInit {
     }
   }
 
+  async showCannotEditActivityDialog() {
+    const popupResult = await this.popupService.showPopup({
```
Julias0, line 1431: Shouldnt we skip the expense if this happens?
can you check what happens in old app in case of activity?

madhavmansuriya40: need to check why we have this on expense page. ideally, this should not be over here
because we can't open expense from the add-edit expense page if I am correct 

but if it is there we can consider it as a `safe to have` check

madhavmansuriya40: this is in the add edit page so it doesn't make sense IMO this won't be executed anytime

PR #157, opened 2021-01-19
Final fixes 32

madhavmansuriya40:


src/app/core/services/recent-local-storage-items.service.ts
```diff
@@ -31,6 +31,10 @@ export class RecentLocalStorageItemsService {
     return recentItems;
   }
 
+  clear(cacheName) {
+    this.storageService.delete(cacheName);
```
Julias0, line 35: this would require async await

Julias0 (approved):

PR #225, opened 2021-02-08
Added last fixes

Julias0:


src/app/fyle/add-edit-per-diem/add-edit-per-diem.page.ts
```diff
@@ -137,6 +137,16 @@ export class AddEditPerDiemPage implements OnInit {
   ngOnInit() {
   }
 
+  get minPerDiemDate() {
+    return this.fg.controls.from_dt.value && moment(this.fg.controls.from_dt.value).subtract(1, 'day').format('y-MM-D');
```
madhavmansuriya40, line 141: i think this might break with double digit date, not sure

iamtarun1993: how this can break sir??? any logic??

madhavmansuriya40: format('y-MM-D') so there was single D so i thought if date will be double digit than it will break but it is Capital D not small D 

this will work, i checked this

madhavmansuriya40 (approved):

iamtarun1993 (approved):

PR #239, opened 2021-02-11
Fyle-1vef81-1: Recently used items service

Dimple16:


Julias0 (approved):

src/app/core/models/recently_used.model.ts
```diff
@@ -0,0 +1,9 @@
+export interface RecentlyUsed {
+    recent_project_ids?: number[];
```
iamtarun1993, line 2: 2 space tabs

src/app/core/services/recently-used-items.service.ts
```diff
@@ -0,0 +1,18 @@
...
+@Injectable({
+  providedIn: 'root'
+})
+export class RecentlyUsedItemsService {
+
+  constructor(
+    private apiService: ApiService
+  ) { }
+
+  getRecentlyUsedV2():Observable<RecentlyUsed> {
```
iamtarun1993, line 15: why this is named as ...v2()
we are not using any v2 api,

iamtarun1993: getRecentlyUsedV2():<space>Observable<RecentlyUsed> {

abhishek1234321 (approved):

src/app/core/services/recently-used-items.service.ts
Dimple16: named it v2 here as in the app/shared-js we are using getRecentlyRsedItems (one from localStorage and other from API) API one is named as v2 - for consistency named the same way here too cc @iamtarun1993

iamtarun1993: Ok, but still will suggest use `getRecentlyUsed` because we are not using any SharedJS here.
@abhishek1234321 @Julias0  need your help.

abhishek1234321: we don't need to name it as v2 here, since sharedjs will not be present here and this is the new angular world so can have the same name in two different cases

abhishek1234321 (changes requested):

src/app/core/services/recently-used-items.service.ts
Dimple16: ^this is resolved cc @iamtarun1993 @abhishek

src/app/core/models/recently_used.model.ts
```diff
@@ -0,0 +1,9 @@
+export interface RecentlyUsed {
+  recent_project_ids?: number[];
+  recent_org_category_ids?: number[];
+  recent_cost_center_ids?: number[];
+  recent_currencies?: string[];
+  recent_vehicle_types?: string[];
+  recent_start_locations?: string[];
+  recent_locations?: string[];
+}
```
abhishek1234321, line 9: add a new line at the end @Dimple16

src/app/core/services/recently-used-items.service.ts
```diff
@@ -0,0 +1,18 @@
...
+export class RecentlyUsedItemsService {
+
+  constructor(
+    private apiService: ApiService
+  ) { }
+
+  getRecentlyUsed():Observable<RecentlyUsed> {
+    return this.apiService.get('/recently_used');
+  }
+}
```
abhishek1234321, line 18: same as above

abhishek1234321 (approved):

PR #244, opened 2021-02-12
Tarun fix 43

iamtarun1993:


src/app/app.component.ts
```diff
@@ -495,6 +496,15 @@ export class AppComponent implements OnInit {
     });
 
     this.setupNetworkWatcher();
+
+    this.router.events.subscribe((ev) => {
+      if (ev instanceof NavigationEnd) {
+        this.menuController.swipeGesture(false);
+        if ((ev.urlAfterRedirects.indexOf('enterprise') > -1) && !(ev.urlAfterRedirects.indexOf('delegated_accounts') > -1)) {
```
Julias0, line 503: what about pre verification and post verification pages?

Julias0, line 503: what about pre verification and post verification pages?

iamtarun1993: no need to show sidemenu on gesture na... enterprise is not present in those pages..
so skipped..

Julias0 (approved):

PR #260, opened 2021-02-18
remove signup flow

madhavmansuriya40:
https://user-images.githubusercontent.com/51825933/109157928-add3b800-7798-11eb-94f0-3241cc1afb6b.mov


src/app/core/services/tracking.service.ts
```diff
@@ -63,17 +40,6 @@ export class TrackingService {
     }
   }
 
-  // external APIs
-  onSignin(email, properties) {
```
Julias0, line 67: why remove onSignin?

Julias0 (changes requested):
Few comments in tracking service. Add a loom video on what happens when app starts up when not logged in and when logged in

src/app/core/services/tracking.service.ts
madhavmansuriya40: by mistake i have done that, but later i added

madhavmansuriya40:
here's loom video link

https://www.loom.com/share/808480ea66de486a9ab49ca4e569c113

src/app/core/services/tracking.service.ts
```diff
@@ -74,15 +60,6 @@ export class TrackingService {
     this.eventTrack('Signin', properties);
   }
 
-  onStateChange(toState, toParams, fromState, fromParams) {
-  }
-
-  eventTrack(action, properties) {
```
Julias0, line 80: why remove this?

src/app/core/services/tracking.service.ts
```diff
@@ -74,15 +60,6 @@ export class TrackingService {
     this.eventTrack('Signin', properties);
   }
 
-  onStateChange(toState, toParams, fromState, fromParams) {
```
Julias0, line 77: confirm once that this is not being used

Julias0 (changes requested):
illa illa

src/app/auth/sign-in/sign-in.page.html
```diff
@@ -75,7 +75,7 @@
     </div>
     <div class="sign-in--redirect">
       Don't have an account?
-      <a [routerLink]="['/', 'auth', 'sign_up']"> Sign Up</a>
+      <a href="https://www.fylehq.com/"> Sign Up</a>
```
abhishek1234321, line 78: do we need `target="_blank"`?

abhishek1234321 (approved):

PR #261, opened 2021-02-18
help page: updated eous/company call to v2 

madhavmansuriya40:


src/app/fyle/help/help.page.ts
```diff
@@ -24,23 +25,30 @@ export class HelpPage implements OnInit {
...
 
   openContactSupportDialog() {
     this.contactSupportLoading = true;
     from(this.loaderService.showLoader('Please wait', 10000)).pipe(
       switchMap(() => {
-        return this.orgUserService.getAllCompanyEouc();
+        return this.authService.getEou();
+      }),
+      switchMap(eou => {
+        return this.orgUserService.getEmployeesByParams({
```
Julias0, line 39: do this inside the service

iamtarun1993 (approved):

src/app/core/services/org-user.service.ts
```diff
@@ -78,6 +80,10 @@ export class OrgUserService {
     );
   }
 
+  getEmployeesByParams(params) {
+    return this.apiV2Service.get('/employees', {params});
```
abhishek1234321, line 84: should we use this or should we use the `search` API, please check with @neeraj once

madhavmansuriya40: here we just need 5 admin users to show in the help dialogue
at other places where to add approver or modify approver functionality is required, we can use search API

abhishek1234321: ok go ahead and merge it in

PR #269, opened 2021-02-22
Changes to make offline expenses deletable

lakshmisuhas1996:


iamtarun1993:
can you put some screenshot or loom video over here?? @lakshmisuhas1996

lakshmisuhas1996:
Loom video - https://www.loom.com/share/e6cad922141c4b00b4a526fb75241352 @iamtarun1993

src/app/fyle/my-expenses/my-expenses.page.ts
```diff
@@ -623,6 +623,23 @@ export class MyExpensesPage implements OnInit {
     }
   }
 
+  async onDeleteOfflineExpenseClick(etxn: Expense, index) {
```
iamtarun1993, line 626: do we need another method to delete offline expenses?
can we not decide based on tx_id is present or not?
if id present make API call, if not present just remove from transactionOutboxService queue.
can you check this?

src/app/fyle/my-expenses/my-expenses-card/my-expenses-card.component.html
```diff
@@ -112,6 +112,13 @@
           <span>Delete</span>
         </div>
       </div>
+
+      <div class="my-expenses-card-menu--buttons" *ngIf="canDeleteOfflineExpense">
```
iamtarun1993, line 116: what will happen if the expense in sync mode?
i.e the user just came online and the expense start syncing to the server. You can slow the network speed and check what is happening here?

iamtarun1993: create 4-5 expense and start syncing and then delete any offline expense, if possible do with images/pdf with all-expense.. it will take more time to sync and give you time to repro.

lakshmisuhas1996: have made changes to not show options icon in expense card when sync is going on...

src/app/fyle/my-expenses/my-expenses.page.ts
```diff
@@ -612,6 +612,9 @@ export class MyExpensesPage implements OnInit {
     if (popupResults === 'primary') {
       from(this.loaderService.showLoader('Deleting Expense', 2500)).pipe(
         switchMap(() => {
+          if (!etxn.tx_id) {
```
iamtarun1993, line 615: Use `iif()` not if else

iamtarun1993: Looks like this will break. deleteOfflineExpense method is a void/promise method but delete() is an observable method.

can you check this once??

iamtarun1993 (approved):

src/app/core/services/transactions-outbox.service.ts
```diff
@@ -227,6 +227,12 @@ export class TransactionsOutboxService {
     return this.dataExtractionQueue;
   }
 
+  deleteOfflineExpense(index: number) {
+    this.queue.splice(index, 1);
+    this.saveQueue()
+    return null;
```
abhishek1234321, line 233: is returning absolutely required? by default the method will return null or undefined no?

src/app/fyle/my-expenses/my-expenses.page.html
```diff
@@ -166,11 +166,12 @@
...
 
   <div *ngFor="let expense of pendingTransactions as list; let i = index">
-    <app-my-expenses-card (addTransactionToNewReport)="onAddTransactionToNewReport($event)" (addTransactionToReport)="onAddTransactionToReport($event)" (deleteTransaction)="onDeleteExpenseClick($event)"
+    <app-my-expenses-card (addTransactionToNewReport)="onAddTransactionToNewReport($event)" (addTransactionToReport)="onAddTransactionToReport($event)" (deleteTransaction)="onDeleteExpenseClick($event, i)"
                           [canOpenCard]="false" [dateComparatorProp]="'tx_txn_dt'"
                           [expense]="expense"
                           [prevExpense]="list[i-1]"
-                          [skipDate]="true">
+                          [skipDate]="true"
+                          [isSyncInProgress]="syncing">
```
abhishek1234321, line 174: isn't `isSyncInProgress` supposed to be a boolean and you are passing a string?

src/app/fyle/my-expenses/my-expenses.page.ts
```diff
@@ -600,7 +600,7 @@ export class MyExpensesPage implements OnInit {
     });
   }
 
-  async onDeleteExpenseClick(etxn: Expense) {
+  async onDeleteExpenseClick(etxn: Expense, index?: number) {
```
abhishek1234321, line 603: why is `index` optional?

abhishek1234321 (changes requested):

src/app/fyle/my-expenses/my-expenses.page.ts
lakshmisuhas1996: SInce we are using the same method to delete normal and offline expenses, we will pass only etxn for normal txns and etxn and index for txns created in offline mode. And delete them based on tx_id prop... so made index as optional param...

src/app/fyle/my-expenses/my-expenses.page.html
lakshmisuhas1996: no, the property **syncing** holds boolean value only - whether the transactions are in syncing or not

src/app/core/services/transactions-outbox.service.ts
lakshmisuhas1996: By default, the return type of function will be Void and it will return undefined. so we are returning null to observable?

abhishek1234321: ok great

abhishek1234321 (approved):

PR #275, opened 2021-02-24
Changes for replacing series of /v2/expenses api call with one stats api call from profile page

lakshmisuhas1996:


src/app/fyle/my-profile/my-profile.page.ts
```diff
@@ -148,14 +148,27 @@ export class MyProfilePage implements OnInit {
...
+    const totalCount = statsCountList.reduce((acc, obj) => acc + obj.value, 0);
+
     return {
-      total: myETxnc.length,
-      mobile: this.transactionService.getCountBySource(myETxnc, 'MOBILE'),
-      extension: this.transactionService.getCountBySource(myETxnc, 'GMAIL'),
-      outlook: this.transactionService.getCountBySource(myETxnc, 'OUTLOOK'),
-      email: this.transactionService.getCountBySource(myETxnc, 'EMAIL'),
-      web: this.transactionService.getCountBySource(myETxnc, 'WEBAPP')
+      total: totalCount,
```
iamtarun1993, line 166: do we need `total:` if we haven't used this... we can remove this

lakshmisuhas1996: This is sorted

iamtarun1993 (approved):

src/app/fyle/my-profile/my-profile.page.ts
```diff
@@ -148,14 +148,27 @@ export class MyProfilePage implements OnInit {
     toast.present();
   }
 
-  setMyExpensesCountBySource(myETxnc) {
+  getMyexpensesStatsCountBySourceMap(stats, source) {
```
abhishek1234321, line 151: what does `BySourceMap` mean?

abhishek1234321, line 151: @Julias0 shouldn't we be specifying type of stats and source? if we do not how will typescript help us catch bugs @lakshmisuhas1996

src/app/fyle/my-profile/my-profile.page.ts
```diff
@@ -148,14 +148,27 @@ export class MyProfilePage implements OnInit {
...
+    let count = stats.filter(stat => stat.key.toLowerCase().indexOf(source.toLowerCase()) > -1);
```
abhishek1234321, line 152: shouldn't this be `const`?

src/app/fyle/my-profile/my-profile.page.ts
abhishek1234321: please explain in code comments about what is happening in this method?

src/app/fyle/my-profile/my-profile.page.ts
```diff
@@ -148,14 +148,27 @@ export class MyProfilePage implements OnInit {
...
+    return count && count.reduce((acc, obj) => acc + obj.value, 0);
```
abhishek1234321, line 153: count will always be an `[]` no? so why do we need to do `count && count.`?

src/app/fyle/my-profile/my-profile.page.ts
```diff
@@ -148,14 +148,27 @@ export class MyProfilePage implements OnInit {
...
+  }
+
+  setMyExpensesCountBySource(stats) {
```
abhishek1234321, line 156: model for `stats`?

src/app/fyle/my-profile/my-profile.page.ts
```diff
@@ -148,14 +148,27 @@ export class MyProfilePage implements OnInit {
...
+    const statsCountList = stats && stats[0].value.map(stat =>  {
```
abhishek1234321, line 157: we shouldn't be doing `stats && stats[0]` the whole point of typescript is to avoid doing things like these, I presume `stats` will always be an array, please define the required model and clean things up? also shouldn't we be checking for `stats.length > 0`? @Julias0 @lakshmisuhas1996 @tarunkumarfyle

src/app/fyle/my-profile/my-profile.page.ts
```diff
@@ -148,14 +148,27 @@ export class MyProfilePage implements OnInit {
...
+      return {
+        value: stat.aggregates[0].function_value,
```
abhishek1234321, line 159: can aggregates ever be empty?

src/app/fyle/my-profile/my-profile.page.ts
```diff
@@ -148,14 +148,27 @@ export class MyProfilePage implements OnInit {
...
+        key: stat.key[0].column_value
```
abhishek1234321, line 160: can key ever be empty?

src/app/fyle/my-profile/my-profile.page.ts
```diff
@@ -148,14 +148,27 @@ export class MyProfilePage implements OnInit {
...
+  }
+
+  setMyExpensesCountBySource(stats) {
+    const statsCountList = stats && stats[0].value.map(stat =>  {
+      return {
+        value: stat.aggregates[0].function_value,
+        key: stat.key[0].column_value  
+      }
+    });
+    const totalCount = statsCountList.reduce((acc, obj) => acc + obj.value, 0);
```
abhishek1234321, line 163: instead of `obj` please use a more readable variable name?

abhishek1234321 (changes requested):

src/app/fyle/my-profile/my-profile.page.ts
Julias0: Yes.

The return type can be skipped in case the value getting returned already has a type set to it.
Ex - 

getEtxn(id: string) {
   return this.apiService.get('/transaction', { id: id }).pipe(
     map(txn => {
        return txn as Transaction;
     })
   )
}

Here typescript would be able to understand the return type of Observable<Transaction>.
But having it explicitly done would be much better

src/app/fyle/my-profile/my-profile.page.ts
Julias0: Yes, ideally the stats model should have a method to retrieve this.
But can't be avoided inside the model since the API return this in this format - properties which are nullable should be marked with ? in the model to have typescript call these out as being optional parameters.

We can enclose these things inside the model itself in the following way - 

ex - 

export class Stats {
  // properties returned by stats here
  
  getStatValue(){
    return this.stats && stats[0] && // etc or use .find to get the relevant stat
  }
}

src/app/fyle/my-profile/my-profile.page.ts
lakshmisuhas1996: yes changing it

src/app/fyle/my-profile/my-profile.page.ts
lakshmisuhas1996: it should be source only not map

abhishek1234321 (approved):

src/app/core/models/stats-one-dimension.model.ts
```diff
@@ -0,0 +1,14 @@
+export interface StatsOneDResponse {
```
Julias0, line 1: add a ? to the params which can be null
Ex - 
export class Person {
  name: string;
  age?: number;
}

? here means the property is nullable

https://stackoverflow.com/questions/17220114/how-to-declare-a-type-as-nullable-in-typescript

Julias0, line 1: Why name this StatsOneResponse?
I mean why is the One there?

src/app/fyle/my-profile/my-profile.page.ts
```diff
@@ -148,14 +149,27 @@ export class MyProfilePage implements OnInit {
...
   }
 
-  setMyExpensesCountBySource(myETxnc) {
+  getMyexpensesStatsCountBySource(stats: any[], source: string) {
+    const filteresStatsRes = stats.filter(stat => stat.key.toLowerCase().indexOf(source.toLowerCase()) > -1);
+    return filteresStatsRes.reduce((acc, statValue) => acc + statValue.value, 0);
+  }
+
+  setMyExpensesCountBySource(statsRes: StatsOneDResponse) {
+    const statsCountList = statsRes.value.map(stat =>  {
```
Julias0, line 158: add it as a method inside the model, then call it here - 

Ex - 
`export class StatsReponse{
  getAggregate(index: number) {
    return this.value.map(stat =>  {
        value: stat.aggregates.length && stat.aggregates[0].function_value,
        key: stat.key.length && stat.key[0].column_value  
    });
  }
}`

Then over here - 
`statsRes.getAggregate(0);`

src/app/fyle/my-profile/my-profile.page.ts
```diff
@@ -148,14 +149,27 @@ export class MyProfilePage implements OnInit {
     toast.present();
   }
 
-  setMyExpensesCountBySource(myETxnc) {
+  getMyexpensesStatsCountBySource(stats: any[], source: string) {
+    const filteresStatsRes = stats.filter(stat => stat.key.toLowerCase().indexOf(source.toLowerCase()) > -1);
+    return filteresStatsRes.reduce((acc, statValue) => acc + statValue.value, 0);
```
Julias0, line 154: This should be a method of the stat model.

src/app/fyle/my-profile/my-profile.page.ts
```diff
@@ -148,14 +149,27 @@ export class MyProfilePage implements OnInit {
     toast.present();
   }
 
-  setMyExpensesCountBySource(myETxnc) {
+  getMyexpensesStatsCountBySource(stats: any[], source: string) {
+    const filteresStatsRes = stats.filter(stat => stat.key.toLowerCase().indexOf(source.toLowerCase()) > -1);
```
Julias0, line 153: `stat.key.toLowerCase().indexOf(source.toLowerCase()) > -1`
This should be a method inside stats

Julias0 (changes requested):
Comments given on requested changes

src/app/core/models/stats-one-dimension.model.ts
lakshmisuhas1996: Named it OneD because the stats res is of type one dimension (stats terminology) and the response structure changes from dimension to dimension. I mean it will be different from 0D and 2D and so on? @Julias0

Julias0: You mean the schema changes when we go from one dimension to multiple dimensions?

lakshmisuhas1996: Yes

src/app/core/models/stats-one-dimension.model.ts
```diff
@@ -0,0 +1,34 @@
...
+  getStatAggregatesByIdx<T>(index: number): Array<{value: T; key: string}> {
+    return this.value.map(stat =>  {
+      return {
+        value: stat.aggregates.length && stat.aggregates[index].function_value,
+        key: stat.key.length && stat.key[index].column_value  
+      }
+    });
+  }
+
+  public static getStatsCountBySource(stats: {value: number; key: string}[], source: string) {
```
Julias0, line 30: dont pass in stats externally.
```
public static getStatsCountBySource(source: string, index = 0) {
  const stats = this.getStatAggregatesByIdx<number>(index);
  const filteresStatsRes = stats.filter(stat => stat.key.toLowerCase().indexOf(source.toLowerCase()) > -1);
  return filteresStatsRes.reduce((acc, statValue) => acc + statValue.value, 0);
}
```

src/app/core/models/stats-one-dimension.model.ts
```diff
@@ -0,0 +1,34 @@
...
+  }];
+  getStatAggregatesByIdx<T>(index: number): Array<{value: T; key: string}> {
+    return this.value.map(stat =>  {
+      return {
+        value: stat.aggregates.length && stat.aggregates[index].function_value,
+        key: stat.key.length && stat.key[index].column_value  
+      }
+    });
+  }
+
```
Julias0, line 29: getStatsTotalCount(index = 0){
  const stats = this.getStatAggregatesByIdx<number>(index);
  return stats.reduce((acc, statValue) => acc + statValue.value, 0);
}

src/app/fyle/my-profile/my-profile.page.ts
```diff
@@ -148,14 +149,17 @@ export class MyProfilePage implements OnInit {
     toast.present();
   }
 
-  setMyExpensesCountBySource(myETxnc) {
+  setMyExpensesCountBySource(statsRes: StatsOneDResponse) {
+    const statsCountList = statsRes.getStatAggregatesByIdx<number>(0);
+    const totalCount = statsCountList.reduce((acc, statValue) => acc + statValue.value, 0);
```
Julias0, line 154: Notice that you dont need statsCountList externally.
You can make it part of the model only

Julias0 (approved):

Julias0 (approved):

Julias0 (approved):

PR #278, opened 2021-02-24
Policy call payload update for new mobile app

Sanyem:


Sanyem:
Created a wrong one :lolsob:

PR #279, opened 2021-02-24
Policy call payload update for new mobile app

Sanyem:


src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -2012,6 +2012,8 @@ export class AddEditExpensePage implements OnInit {
           }
         }
 
+        policyETxn.tx.corporate_credit_card_expense_group_id = !!this.selectedCCCTransaction;
```
iamtarun1993, line 2015: Do you need `policyETxn.tx.corporate_credit_card_expense_group_id` value as a boolean only??

iamtarun1993: If this is the case then LGTM.

iamtarun1993 (approved):

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
Sanyem: Yup; need this as boolean only

abhishek1234321: hey, not understand this, `corporate_credit_card_expense_group_id` this is supposed to be id and not boolean no? this is like saying I will use `tx_id` but pass a boolean

Sanyem: Hey, there's a reason behind this - Ideally this should be an id and not a boolean but while creating new expenses, this id is not yet computed or stored in db but we make policy /test calls and we have to violate the policy incase the expense is not matched. That's why for policy /test call we are sending this as boolean. For policy /result call that is made from eventhandler, we are sending this key as id. 

We are using the same key name as both policy calls use the same condition validator and it would be hacky to use different key names for different calls. Also, for policy /result call we modify the etxn before checking the conditions to make this key into a boolean and not specific ids. 

We are not concerned by what id is present. We are only concerned that some id should be present for an expense to be matched

abhishek1234321: it should be the policy code's job to identify whether it is present or not and then convert to boolean, in the present scenario every client has to do this no? I am happy to get on a call if need be

abhishek1234321 (changes requested):

abhishek1234321 (approved):

iamtarun1993 (approved):

PR #290, opened 2021-03-04
Fix for freshdesk chat in ios safearea rendering

iamtarun1993:


iamtarun1993:
New -> 
Old ->

abhishek1234321:
you are hiding time no @tarunkumarfyle ? is this what is followed everywhere? it looks a bit odd

src/global.scss
```diff
@@ -423,6 +423,10 @@ input[type="date"]::after {
 
 #fc_frame {
   display: none;
+  padding-top: constant(safe-area-inset-top); /*for iOS 11.2 and above*/
```
abhishek1234321, line 426: does ionic not handle this out of the box? or is it because we use freshchat, we have to take care of this?

abhishek1234321 (changes requested):

src/global.scss
iamtarun1993: ionic handle its own component... but it can't handle any frame created by other libraries.

so we need to take care of this..

iamtarun1993:
> you are hiding time no @tarunkumarfyle ? is this what is followed everywhere? it looks a bit odd

no sir...
It was hidden in the old app.. but showing in the new app.

abhishek1234321 (approved):

PR #293, opened 2021-03-08
Added cost center Model

lakshmisuhas1996:
- Added CostCenter Model
- Changes for to use CostCenter Model in cost-center service


src/app/core/models/cost-center.model.ts
```diff
@@ -0,0 +1,10 @@
+export interface CostCenter {
+  id: number;
+  created_at: Date;
+  updated_at: Date;
+  name: string;
+  code?: string;
+  active?: boolean;
```
madhavmansuriya40, line 7: active has a default value in DB, can remove the `?`

madhavmansuriya40 (approved):
resolve comments and assign further in pipeline

PR #298, opened 2021-03-09
Geo location fixes

iamtarun1993:


src/app/core/services/location.service.ts
```diff
@@ -93,4 +98,15 @@ export class LocationService {
       })
     );
   }
+
+  @Cacheable({
+    cacheBusterObserver: currentLocationCacheBuster$,
+    maxAge: 10 * 60 * 1000 //10 minute
```
abhishek1234321, line 104: `// 10 minutes`

src/app/core/services/location.service.ts
```diff
@@ -93,4 +98,15 @@ export class LocationService {
...
+  })
+  getCurrentLocation(enableHighAccuracy?: boolean) {
```
abhishek1234321, line 106: why is this optional `enableHighAccuracy`?

src/app/core/services/location.service.ts
```diff
@@ -93,4 +98,15 @@ export class LocationService {
...
+    return from(Geolocation.getCurrentPosition({
+      timeout: 10000,
```
abhishek1234321, line 108: keep 5s timeout, 10 is a lot

abhishek1234321 (approved):

src/app/core/services/location.service.ts
Julias0: @tarunkumarfyle lets give this a default value of value than make it optional.

iamtarun1993: We can do mandatory also.. but no issue with optional also. If no value is passing from the method it will consider it as false. 
will change this to mandatory

src/app/core/services/location.service.ts
iamtarun1993: @Julias0 if we reduce it here, we need to reduce it in this place also an
`this.loaderService.showLoader('Loading current location...', 10000)`

and any other place??

src/app/core/services/location.service.ts
Julias0: Also add a return type

Julias0 (approved):

PR #299, opened 2021-03-10
Trips delete, pull back, close button disable when api call in progress

madhavmansuriya40:


iamtarun1993 (approved):

src/app/fyle/my-view-trips/my-view-trips.page.ts
```diff
@@ -53,6 +53,9 @@ export class MyViewTripsPage implements OnInit {
   canCloseTrip$: Observable<boolean>;
   canDelete$: Observable<boolean>;
   canEdit$: Observable<boolean>;
+  pullbackLoading = false;
```
abhishek1234321, line 56: do we really need different variables for different loaders? can we not use one?

src/app/fyle/my-view-trips/my-view-trips.page.ts
```diff
@@ -74,6 +77,10 @@ export class MyViewTripsPage implements OnInit {
 
 
   async deleteTrip() {
```
abhishek1234321, line 79: when is this method called? any screenshots or loom will really help

abhishek1234321 (changes requested):

src/app/fyle/my-view-trips/my-view-trips.page.ts
madhavmansuriya40: if we use 1, with all button
the loader will be shown in all buttons at the same time

src/app/fyle/my-view-trips/my-view-trips.page.ts
madhavmansuriya40: ok will add loom video link

src/app/fyle/my-view-trips/my-view-trips.page.ts
madhavmansuriya40: sir i checked about `using 1 variable for showing loader in all buttons` i think this can be done but that will be tricky and hard work, also need to update at many places and all behaviours will not be same

madhavmansuriya40: we can have a separate task for refactoring `loaders inside button` component

abhishek1234321 (approved):

PR #311, opened 2021-03-15
API call for user_app_version should not go if device is not present from browser

lakshmisuhas1996:


src/app/core/services/offline.service.ts
```diff
@@ -62,7 +64,11 @@ export class OfflineService {
     const homeCurrency$ = this.getHomeCurrency();
     const delegatedAccounts$ = this.getDelegatedAccounts();
 
-    this.appVersionService.load();
+    from(this.deviceService.getDeviceInfo()).subscribe(deviceInfo => {
```
madhavmansuriya40, line 67: `this.deviceService.getDeviceInfo()` returns an observable can't we directly subscribe to that ?
I think in that case we will not need `from()`

madhavmansuriya40 (approved):

src/app/core/services/offline.service.ts
```diff
@@ -62,7 +64,11 @@ export class OfflineService {
     const homeCurrency$ = this.getHomeCurrency();
     const delegatedAccounts$ = this.getDelegatedAccounts();
 
-    this.appVersionService.load();
+    this.deviceService.getDeviceInfo().subscribe(deviceInfo => {
+      if (deviceInfo.platform === 'ios' || deviceInfo.platform === 'android') {
```
iamtarun1993, line 68: deviceInfo.platform.toLowerCase() === 'ios'

iamtarun1993 (changes requested):
use .toLowerCase()

iamtarun1993 (approved):

abhishek1234321 (approved):

PR #314, opened 2021-03-15
Fyle t 2dd988 notch fix

iamtarun1993:


src/app/fyle/add-edit-expense/add-edit-expense.page.html
```diff
@@ -831,28 +831,30 @@
 
 <ng-container *ngIf="fg">
   <ion-footer>
```
iamtarun1993, line 833: just need to remove the class from `ion-toolbar` and put the class inside an extra `div`

src/global.scss
```diff
@@ -83,6 +83,15 @@ input:focus, textarea:focus{
 
   .popover-viewport {
```
iamtarun1993, line 84: need to fix it globally, else we need to fix it in every modal and popover individually.

abhishek1234321 (approved):

PR #318, opened 2021-03-17
User Service Model changes

lakshmisuhas1996:


src/app/core/models/user-property.model.ts
```diff
@@ -0,0 +1,6 @@
+export interface UserProperty {
+  id?: string;
```
lakshmisuhas1996, line 2: keeping all properties optional because for some users api res return empty object does not have any property in it.

madhavmansuriya40 (approved):
please also check if we user user service methods at other places and do required changes

src/app/core/models/user-property.model.ts
```diff
@@ -0,0 +1,6 @@
...
+  created_at?: Date;
+  updated_at?: Date;
+  devices?: { id: string; fcm_token: string; }[];
```
iamtarun1993, line 5: create another interface and use the new model here

src/app/core/models/user-property.model.ts
iamtarun1993: how this can be possible.. it will always return `id, create_at, updated_at` 
`devices` can be an empty array.

iamtarun1993: check with `tarung.kumar@fyle.in`

iamtarun1993 (changes requested):
some major comments

src/app/core/models/user-property.model.ts
lakshmisuhas1996: For yours (u_id - usHOvyUSnZPS) was getting response with all props and for my acc (u_id - us2t5LrjG3sk) not getting any res. And also in **push-notification.service.ts** we are checking the res came is null or not?

iamtarun1993: ok.. got it, looks like this is a bug from the backend. let's keep optional thingy.

src/app/core/models/user-property.model.ts
```diff
@@ -0,0 +1,11 @@
+export interface DeviceInfo {
```
iamtarun1993, line 1: move this to under `V1` directory

src/app/core/models/v1/user-property.model.ts
```diff
@@ -0,0 +1,11 @@
+export interface DeviceInfo {
+  id: string;
+  fcm_token: string;
+}
+
+export interface UserProperty {
+  id?: string;
```
abhishek1234321, line 7: not understanding this, can all the properties be optional? isn't that odd?

abhishek1234321 (changes requested):

src/app/core/models/v1/user-property.model.ts
lakshmisuhas1996: For this API call **users/us_id/properties**, for some user accounts the response has all those properties in the response and for some the response was empty. And also in push-notification.service (line 71) we are checking if the res is empty or not and upserting user properties if not present

abhishek1234321: is this optional in API, if not then why keep this optional?

abhishek1234321 (changes requested):

abhishek1234321 (approved):

PR #322, opened 2021-03-18
ReceiptService Model changes

lakshmisuhas1996:


src/app/core/models/receipt.model.ts
```diff
@@ -0,0 +1,26 @@
+import { OrgUser } from './org-user.model';
+
+export interface ExtractedData {
+  amount: number;
```
madhavmansuriya40, line 4: all the data will be present every time?

there are many different extraction services (different for PDF, image, HTML), not sure but please check once

src/app/core/services/receipt.service.ts
```diff
@@ -31,7 +34,11 @@ export class ReceiptService {
     const data = {
       params
     };
-    return this.apiv2Service.get('/receipts', data);
+    return this.apiv2Service.get('/receipts', data).pipe(
+      map(
```
madhavmansuriya40, line 38: `map(res => res.data[0] as Receipt)`
 
make it 1 liner

PR #323, opened 2021-03-18
Fix for iOS signup issue

iamtarun1993:


src/app/auth/sign-in/sign-in.page.html
```diff
@@ -73,9 +73,9 @@
       SIGN IN WITH GOOGLE
     </button>
     </div>
-    <div class="sign-in--redirect">
+    <div class="sign-in--redirect" (click)="signUp()">
```
abhishek1234321, line 76: `goToFyleWebsite`

abhishek1234321 (approved):

PR #343, opened 2021-03-26
Fyle filter pipe try1 - Trying to implement filter pipe

shishirams222:


src/app/shared/pipes/filter.pipe.ts
```diff
@@ -0,0 +1,25 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'filter'
+})
+
+export class FilterPipe implements PipeTransform {
+  transform(items: any[], searchText: string): any[] {
```
abhishek1234321, line 8: would it be `any[]` always or can we be more specific? like `string[]` etc?

src/app/shared/pipes/filter.pipe.ts
```diff
@@ -0,0 +1,25 @@
...
+    // if there are no items, return empty list
+		if (!items) {
```
abhishek1234321, line 10: indentation 2 spaces please

abhishek1234321, line 10: what if `items` is an `[]`?

src/app/shared/pipes/filter.pipe.ts
```diff
@@ -0,0 +1,25 @@
...
+			return [];
+		}
+
+		// if there is no search text, return full list
+		if (!searchText) {
```
abhishek1234321, line 15: do we have any convention wherein some parameters can be optional in a filter?

src/app/shared/pipes/filter.pipe.ts
```diff
@@ -0,0 +1,25 @@
...
+			return items; 
+		}
+
+		searchText = searchText.toLowerCase();
```
abhishek1234321, line 19: searchTextLowerCase

src/app/shared/pipes/filter.pipe.ts
```diff
@@ -0,0 +1,25 @@
...
+
+		return items.filter ( item => {
+      return item.toLowerCase().includes(searchText);
```
abhishek1234321, line 22: indentation is off

abhishek1234321 (changes requested):

src/app/shared/pipes/filter.pipe.ts
shishirams222: It should be `any[]`, list might not be just list of strings. So kept it as `any[]`. There are more checks inside

src/app/shared/pipes/filter.pipe.ts
shishirams222: Updated

shishirams222: updated

src/app/shared/pipes/filter.pipe.ts
```diff
@@ -0,0 +1,34 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'filter'
+})
+
```
Julias0, line 6: no empty line please

PR #356, opened 2021-04-01
Return null instead of throwError 

iamtarun1993:


madhavmansuriya40 (approved):

madhavmansuriya40 (approved):

Julias0 (approved):

src/app/core/services/permissions.service.ts
```diff
@@ -2229,7 +2229,7 @@ export class PermissionsService {
           if (currentAllowedActions.allowedRouteAccess) {
             return of(currentAllowedActions);
           } else {
-            throwError('no route access');
+            return of(null);
```
Julias0, line 2232: shouldnt we return the throwError?

iamtarun1993: If we return `throwError` also then the sentry also catch this and put in their console. So no point of solving this

Julias0: But why is this error coming in the first place?

src/app/core/services/permissions.service.ts
```diff
@@ -2202,7 +2202,7 @@ export class PermissionsService {
         }
 
         return roles;
-      })
+      }),filter(roles => roles.length > 0)
```
Julias0, line 2205: lets skip this approach

src/app/core/services/permissions.service.ts
```diff
@@ -2202,7 +2202,7 @@ export class PermissionsService {
...
     );
 
     return filteredRoles$.pipe(
```
Julias0, line 2208: use iif to check if fitleredRoles is null or empty array. If yes, return of(null) else lets go with this flow.

Using filter will prevent this observable from resolving which might have unintented consequences

src/app/core/services/permissions.service.ts
```diff
@@ -2229,11 +2229,17 @@ export class PermissionsService {
           if (currentAllowedActions.allowedRouteAccess) {
             return of(currentAllowedActions);
           } else {
-            throwError('no route access');
+            return throwError('no route access');
```
abhishek1234321, line 2232: is this an observable?

abhishek1234321, line 2232: is this an observable?

abhishek1234321, line 2232: is this an observable?

abhishek1234321, line 2232: is this an observable?

abhishek1234321 (changes requested):

abhishek1234321 (approved):

src/app/core/services/permissions.service.ts
iamtarun1993: yes throwErrror always return observables

Julias0 (approved):

PR #362, opened 2021-04-05
Changes for date input value is overlapping with formatted value

lakshmisuhas1996:


src/global.scss
```diff
@@ -458,12 +458,14 @@ ref - https://stackoverflow.com/a/31162426
...
   &__format{
     position: relative;
+    color: transparent;
 
     &::before {
       position: absolute;
       left: 4px;
       content: attr(data-date);
       display: inline-block;
+      color: black;
```
iamtarun1993, line 468: color: #000;

src/global.scss
```diff
@@ -458,12 +458,14 @@ ref - https://stackoverflow.com/a/31162426
...
   &__format{
     position: relative;
+    color: transparent;
 
     &::before {
       position: absolute;
       left: 4px;
       content: attr(data-date);
       display: inline-block;
+      color: #000;
```
abhishek1234321, line 468: please use a SCSS variable

src/global.scss
```diff
@@ -458,12 +458,14 @@ ref - https://stackoverflow.com/a/31162426
 .date-input {
   &__format{
     position: relative;
+    color: transparent;
```
abhishek1234321, line 461: can we use a hex code here?

lakshmisuhas1996: Have tried with rgb & rgba codes here.. Only setting color to transparent and white working here?

abhishek1234321 (approved):

PR #364, opened 2021-04-06
Fyle inverse highlight: Text highlight changes for fy-select dropdowns

shishirams222:
Screenshots:
Screenshot 2021-04-06 at 1 28 58 PM
Screenshot 2021-04-06 at 1 25 14 PM


src/global.scss
```diff
@@ -449,3 +449,10 @@ input[type="date"]::after {
 .button-outline {
   --border-width: 1px;
 }
+
+.inverse-highlight {
```
Julias0, line 453: move this to component ka scss file

src/app/shared/pipes/highlight.pipe.ts
```diff
@@ -0,0 +1,25 @@
+import {Pipe, PipeTransform} from '@angular/core';
+
+// ref: https://forum.ionicframework.com/t/highlighting-text/105265/2
+@Pipe({name: 'highlight'})
+export class HighlightPipe implements PipeTransform {
+  transform(text: string, search: string): string {
+    try {
+      if ((text && text.length > 0) && (search && search.length > 0)) {
+        text = text.toString(); // sometimes comes in as number
```
abhishek1234321, line 9: when does it come as a number? if it comes as a number and you return a string, wouldn't it cause other issues? when the selection happens in the modal and stuff?

src/app/shared/pipes/highlight.pipe.ts
```diff
@@ -0,0 +1,25 @@
...
+        search = search.trim();
+        if (search.length > 0) {
+          let pattern = search.trim().replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&');
```
abhishek1234321, line 12: what does this do? explain in code comments and give SO ref in code comments

src/app/shared/pipes/highlight.pipe.ts
```diff
@@ -0,0 +1,25 @@
...
+          pattern = pattern.split(' ').filter((t) => {
+              return t.length > 0;
+          }).join('|');
```
abhishek1234321, line 15: why join with `|`, explain with an example in code comments

abhishek1234321 (changes requested):

src/app/shared/pipes/highlight.pipe.ts
shishirams222: There no cases currently where it will be a number. So removed it. Changes in the existing places are working fine post removal

src/app/shared/pipes/highlight.pipe.ts
shishirams222: Added reference and explained in detail with a example for every step

src/app/shared/pipes/highlight.pipe.ts
shishirams222: Updated

Julias0 (approved):

abhishek1234321 (approved):

PR #367, opened 2021-04-08
Only one access_toke call and show loader in sign_in page when user close and open app after sign in

iamtarun1993:


src/app/auth/sign-in/sign-in.page.ts
```diff
@@ -252,17 +252,20 @@ export class SignInPage implements OnInit {
     this.emailSet = !!this.fg.value.email;
   }
 
-  async ngOnInit() {
+  ngOnInit() {
     const presentEmail = this.activatedRoute.snapshot.params.email;
     this.fg = this.formBuilder.group({
       email: [presentEmail || '', Validators.compose([Validators.required, Validators.pattern('\\S+@\\S+\\.\\S{2,}')])],
       password: ['', Validators.required]
     });
```
iamtarun1993, line 261: Converted this promise-based logic into observable logic. And showing one loader when checking the user is already logged in or not.

src/app/core/interceptors/httpInterceptor.ts
```diff
@@ -47,19 +41,6 @@ export class HttpConfigInterceptor implements HttpInterceptor {
     return false;
   }
```
iamtarun1993, line 42: moved these two methods into `routerAuthService`.

lakshmisuhas1996 (approved):

abhishek1234321 (approved):

src/app/core/interceptors/httpInterceptor.ts
```diff
@@ -10,19 +10,21 @@ import {
...
 import * as moment from 'moment';
 import { TokenService } from '../services/token.service';
-import { AuthService } from '../services/auth.service';
 import { RouterAuthService } from '../services/router-auth.service';
 import { DeviceService } from '../services/device.service';
 
 @Injectable()
 export class HttpConfigInterceptor implements HttpInterceptor {
+  private accessTokenCallInProgress = false;
+  private accessTokenSubject: Subject<any> = new BehaviorSubject<any>(null);
```
Julias0, line 26: private accessTokenSubject = new BehaviorSubject<any>(null);

Also replace any with proper type

src/app/core/interceptors/httpInterceptor.ts
```diff
@@ -72,21 +73,44 @@ export class HttpConfigInterceptor implements HttpInterceptor {
...
-            concatMap(authResponse => {
-              return from(this.routerAuthService.newAccessToken(authResponse.access_token))
-            }),
-            concatMap(() => {
-              return from(this.tokenService.getAccessToken())
-            })
-          );
+          if (!this.accessTokenCallInProgress) {
+            this.accessTokenCallInProgress = true;
+            this.accessTokenSubject.next(null);
```
Julias0, line 88: why are we emitting null here?

src/app/core/interceptors/httpInterceptor.ts
```diff
@@ -72,21 +73,44 @@ export class HttpConfigInterceptor implements HttpInterceptor {
...
+            return from(this.tokenService.getRefreshToken()).pipe(
+              concatMap(refreshToken => {
```
Julias0, line 90: use switchMap instead of concat map

src/app/core/interceptors/httpInterceptor.ts
```diff
@@ -72,21 +73,44 @@ export class HttpConfigInterceptor implements HttpInterceptor {
...
   }
 
-  getAccessToken() {
+  /**
+   * This method get current accessToken from Storage, check if this token is expiring or not.
+   * If the token is expiring it will get another accessToken from API and return the new accessToken
+   * If multiple API call initiated then `this.accessTokenCallInProgress` will block multiple access_token call
+   * Reference: https://stackoverflow.com/a/57638101
+   */
+   getAccessToken(): Observable<string> {
```
Julias0, line 82: if we are using behaviorSubject

this should be 
```
getAccessToken(): Observable<string>{
  return  this.refreshAccessTokenIfExpired.pipe(
    switchMap(()=> this. accessTokenSubject.asObservable())
   );
}
```

the flag `accessTokenCallInProgress` shouldnt be required - accessTokenSubject should only emit when required

iamtarun1993: check this 
https://stackoverflow.com/a/57638101

src/app/core/interceptors/httpInterceptor.ts
Julias0: dei this is not taken care of

src/app/core/interceptors/httpInterceptor.ts
iamtarun1993: copied from https://stackoverflow.com/a/57638101

Julias0 (approved):

PR #368, opened 2021-04-08
Switch to org option should not be visible when switched to delegatee account

lakshmisuhas1996:


iamtarun1993 (approved):

src/app/app.component.ts
```diff
@@ -319,7 +319,7 @@ export class AppComponent implements OnInit {
           },
           {
             title: 'Switch Organization',
-            isVisible: (orgs.length > 1),
+            isVisible: (orgs.length > 1) && !this.isSwitchedToDelegator,
```
abhishek1234321, line 322: if I am in the delegatee's account, how do I switch to other delegatee's account?

abhishek1234321 (changes requested):

src/app/app.component.ts
lakshmisuhas1996: When we switch to any delegatee's account, we will get an option to **Switch back to own account** then we can switch to other delegatee's account by clicking on **Delegated Accounts**
Screen Shot 2021-04-09 at 10 01 44 AM
Screen Shot 2021-04-09 at 10 03 44 AM

abhishek1234321 (approved):

PR #373, opened 2021-04-12
Changing Logo header background same as menu bg color in side menu

lakshmisuhas1996:


src/app/app.component.scss
```diff
@@ -25,8 +24,8 @@ $sidenav-bg: #161528;
     position: relative;
     height: 100%;
     width: 100%;
-    --background: #111023;
-    border-color: $header-bg;
+    --background: #161528;
```
lakshmisuhas1996, line 27: Have tried using a variable here but changes are not reflecting...same with changing **--background** property to **background**

iamtarun1993: check this
https://ionicframework.com/docs/theming/css-variables#using-css

iamtarun1993 (approved):

Julias0 (approved):

PR #387, opened 2021-04-28
Save extracted_data in transaction POST object(Already fixed by Aiyush in another branch)

iamtarun1993:


src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -2925,6 +2926,8 @@ export class AddEditExpensePage implements OnInit {
         invoice_dt: imageData && imageData.data && imageData.data.invoice_dt || null
       };
 
+      etxn.tx.extracted_data = extractedData;
```
Julias0, line 2929: we should do this when the generated etxn is created.
This is storing data inside an observable which is wrong

PR #410, opened 2021-05-28
Logout user when access_token call is failed

iamtarun1993:


src/app/core/interceptors/httpInterceptor.ts
```diff
@@ -66,6 +71,12 @@ export class HttpConfigInterceptor implements HttpInterceptor {
       concatMap(
         refreshToken => this.routerAuthService.fetchAccessToken(refreshToken)
       ),
+      catchError(error => {
+        this.userEventService.logout();
```
Julias0, line 75: dosnt 
```
this.storageService.clearAll();        
globalCacheBusterNotifier.next();
```
happen on the handler of `this.userEventService.logout();` ?

iamtarun1993: no these 2 is not happening in this handler. We need to do explicitly, same in my profile page.

Julias0: Interesting - we should clean this up otherwise would violate DRY. I am ok with doing this separatly

Julias0 (approved):

PR #429, opened 2021-06-14
Added changes for search

Julias0:


Julias0:
Screenshot 2021-06-16 at 1 10 07 PM
Screenshot 2021-06-16 at 1 10 18 PM
Screenshot 2021-06-16 at 1 10 26 PM
Screenshot 2021-06-16 at 1 08 45 PM
Screenshot 2021-06-16 at 1 08 54 PM
Screenshot 2021-06-16 at 1 09 11 PM
Screenshot 2021-06-16 at 1 09 20 PM

src/app/shared/components/fy-filters/fy-filters.component.ts
```diff
@@ -0,0 +1,122 @@
...
+  get FilterOptionType() {
+    return FilterOptionType;
+  }
+
+  constructor(
+      private modalController: ModalController
+  ) { }
+
+  ngOnInit() {
+    const activeFilterInitialIndex = (this.activeFilterInitialName && this.filterOptions.findIndex(option => option.name === this.activeFilterInitialName)) || 0;
```
abhishek1234321, line 36: why default to 0?

abhishek1234321: is it to select the first element?

src/app/shared/components/fy-filters/fy-filters.component.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  onDateChange() {
+    this.customDateMap[this.activeFilter.name] = {
+      startDate: this.startDate,
+      endDate: this.endDate
+    };
+  }
+
+  switchFilter(currentFilter: FilterOptions<any>, option: { label: string; value: any }) {
```
abhishek1234321, line 73: what is `switchFilter` about?

src/app/shared/components/fy-filters/selected-filters.interface.ts
```diff
@@ -0,0 +1,8 @@
+export interface SelectedFilters<T> {
+  name: string;
+  value: T | T[];
```
abhishek1234321, line 3: is this a good way? sometimes `string` sometimes `[]` of numbers :P

src/app/shared/components/fy-filters/selected-filters.interface.ts
```diff
@@ -0,0 +1,8 @@
...
+  associatedData?: {
```
abhishek1234321, line 4: you could have this without `?` and use `as` this usually works well when you use things for `post` and you don't have `id`

abhishek1234321 (changes requested):

src/app/shared/components/fy-filters/selected-filters.interface.ts
Julias0: `as` is used for typecasting data but since this is an interface we cant use it here

src/app/shared/components/fy-filters/selected-filters.interface.ts
Julias0: Using unions is actually ok in typescript - since typescript doesn't have a good way of serialization of classes, we cant do it in the object-oriented way of creating implementations of the interface and only having methods here.

Heres a reference why unions are awesome - https://reallifeprogramming.com/why-union-is-your-best-friend-in-modeling-types-in-typescript-bb6d53a130ba

src/app/shared/components/fy-filters/fy-filters.component.ts
Julias0: switch filter is about maintaining the state of the filter which is currently being modified by type.

src/app/shared/components/fy-filters/fy-filters.component.ts
Julias0: Yes, if no activeFilterName is passed we default to the first filter as the activeFilter

src/app/shared/components/fy-filters/selected-filters.interface.ts
Julias0: `as` is used for typecasting data mostly - not for defining types 
https://stackoverflow.com/questions/55781559/what-does-the-as-keyword-do

src/app/shared/components/fy-filters/fy-filters.component.ts
```diff
@@ -0,0 +1,122 @@
...
+  templateUrl: './fy-filters.component.html',
+  styleUrls: ['./fy-filters.component.scss'],
+})
+export class FyFiltersComponent implements OnInit {
+  @Input() filterOptions: FilterOptions<any>[];
+  @Input() selectedFilterValues: SelectedFilters<any>[];
+  @Input() activeFilterInitialName;
+
+  currentFilterValueMap: {[key: string]: any| any[]} = {};
+  customDateMap: {[key: string]: {
```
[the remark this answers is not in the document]
Julias0: we would have to use `?` or Partial with `as` since the params would still be required . Plus `as` is used for typecasting mostly

src/app/shared/components/fy-filters/fy-filters.component.html
```diff
@@ -0,0 +1,112 @@
...
+    </ion-row>
+  </ion-grid>
+</ion-content>
+<ion-footer>
+  <ion-toolbar>
+    <div class="fy-filters--footer-toolbar">
+      <ion-button class="fy-filters--cancel" (click)="cancel()" fill="clear">
+        Cancel
+      </ion-button>
+      <ion-button class="fy-filters--save" (click)="save()" expand="block" color="fyle-primary" fill="solid">
```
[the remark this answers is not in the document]
Julias0: ionic defaults this to `button` 
https://ionicframework.com/docs/api/button
I don't think we should use type='submit' since that is used in old style forms and I would have to add a preventDefault

src/app/shared/components/fy-filters/fy-filters.component.scss
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  &--grid {
+    --ion-grid-padding: 0;
+    --ion-grid-column-padding: 0;
+    height: 100%;
+  }
+
+  &--row {
+    height: 100%;
```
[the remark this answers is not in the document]
Julias0: Yes, the row optimizes space on what is inside but in this case we can to expand the space to take up the whole space inside the body of the filter

src/app/shared/components/fy-filters/fy-filters.component.scss
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  &--filter-item {
+    --padding-start: 0;
+    --inner-padding-start: 16px;
+    --inner-padding-top: 13px;
```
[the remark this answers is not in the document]
Julias0: the ion label which is used inside has a padding of 9px by default. This is basically to sum it up and make total padding 22px

src/app/shared/components/fy-filters/fy-filters.component.html
```diff
@@ -0,0 +1,112 @@
...
+            <ng-template #uncheckedSSRadioBtn>
+              <ion-icon class="fy-filters--radio" src="../../../../assets/svg/fy-radio-unchecked.svg"></ion-icon>
+            </ng-template>
+            {{option.label}}
+          </ion-item>
+          <ng-container *ngIf="currentFilterValueMap[activeFilter.name] === 'custom'">
+            <div class="fy-filters--custom-date-filters">
+              <mat-form-field>
+                <mat-label>start date</mat-label>
+                <input matInput [(ngModel)]="startDate" (ngModelChange)="onDateChange()" [matDatepicker]="picker1">
```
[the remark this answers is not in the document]
Julias0: picker1 is the reference to `<mat-datepicker #picker1></mat-datepicker>` 
This is required for the input to link to the datepicker - has no usage / will have no usage otherwise.

src/app/shared/components/fy-filters/filter-options.interface.ts
```diff
@@ -0,0 +1,7 @@
+import {FilterOptionType} from './filter-option-type.enum';
+
+export interface FilterOptions<T> {
+  optionType: FilterOptionType;
+  name: string;
+  options: { label: string, value: T }[];
```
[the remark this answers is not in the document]
Julias0: Since filters is generic the value can be anything - strings, enums etc. Plus this is not getting it at runtime - this would be creating a generic class that is resolved at compile-time only.

src/app/shared/components/fy-delete-dialog/fy-delete-dialog.component.ts
```diff
@@ -0,0 +1,52 @@
...
+})
+export class FyDeleteDialogComponent implements OnInit {
+
+  @Input() deleteMethod: () => Observable<any>;
+  @Input() header: string;
+  @Input() body: string;
+
+  deleteCallInProgress = false;
+
+  get LoaderPosition() {
```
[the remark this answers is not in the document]
Julias0: This is to get the enum definition available inside the template via a getter -
https://stackoverflow.com/questions/59289095/use-of-enums-in-angular-8-html-template-for-ngif

PR #430, opened 2021-06-15
Capacitor upgrade

Julias0:


src/app/auth/new-password/popup/popup.component.spec.ts
```diff
@@ -1,24 +0,0 @@
-import {async, ComponentFixture, TestBed} from '@angular/core/testing';
```
Julias0, line 1: ignore

Julias0:
Add node version to README

PR #436, opened 2021-06-19
Upgrade Ionic, Angular and its dependencies

abhishek1234321:


angular.json
```diff
@@ -33,13 +33,22 @@
...
               }
             ],
-            "scripts": []
+            "scripts": [],
+            "aot": false,
+            "vendorChunk": true,
+            "extractLicenses": false,
+            "buildOptimizer": false,
+            "sourceMap": true,
+            "optimization": false,
```
iamtarun1993, line 50: optimization should be true IIRC, else size of bundle will be high

abhishek1234321: can you confirm? this came when I did `ng update` so I am not so sure

PR #439, opened 2021-06-22
FYLE-3vfnz6 Fix Default currency settings option

iravisankar:
screenshot-nimbus-capture-2021 06 22-18_09_05
screenshot-nimbus-capture-2021 06 22-18_07_58


iamtarun1993:
If you are fixing a bug in UI, please always put one screenshot before fix and one after the fix.
Without these screenshots, it's hard to guess.

iamtarun1993 (approved):

PR #443, opened 2021-06-22
FYLE-3xb6tg Added ESLint for linting Support

iravisankar:


.eslintrc.json
```diff
@@ -0,0 +1,61 @@
+{
```
iamtarun1993, line 1: in old tslint.json files we have rules like 2 space tab, that is missing here?
does tslint apply 2 space tab automatically?

iamtarun1993: I also can't check any files got changed because of this eslint new rules.
IIRC many files should be changed.

PR #485, opened 2021-07-16
FYLE-49fazr: Zero state alignment fixed in expenses page

Dimple16:
**Iphone X:**
Screenshot 2021-07-16 at 12 57 30 PM

### DIFF
--- a/src/app/fyle/my-expenses/my-expenses.page.html
+++ b/src/app/fyle/my-expenses/my-expenses.page.html
@@ -63,7 +63,7 @@
 </app-fy-header>
 
 <ion-content class="my-expenses--content">
-  <div class="my-expenses--body">
+  <div class="my-expenses--body" [ngClass]="{'my-expenses--zero-states-body': ((count$ | async) === 0) && isConnected$|async}">
   <ion-refresher (ionRefresh)="doRefresh($event)" *ngIf="isConnected$|async" slot="fixed">
     <ion-refresher-content></ion-refresher-content>
   </ion-refresher>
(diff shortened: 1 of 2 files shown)


src/app/fyle/my-expenses/my-expenses.page.html
```diff
@@ -96,10 +96,6 @@
       </ng-template>
     </div>
```
Dimple16, line 98: Removed below action button as it is not valid anymore due to bottom navigation camera icon

src/app/fyle/my-expenses/my-expenses.page.scss
```diff
@@ -307,6 +265,7 @@
   &--body {
     padding-bottom: 20px;
     background-color: $white;
+    height: 100%;
```
iamtarun1993, line 268: IIRC I removed this because of some issue. can you check this?
1. Last expense is visible properly on the iPhone X screen, and there is a small gap b/w the camera start and expense end?
2. During scroll infinite scroll is visible properly?

src/app/fyle/my-expenses/my-expenses.page.html
iamtarun1993: Yes, I also removed this in my private branch.

src/app/fyle/my-expenses/my-expenses.page.scss
Dimple16: added `height: 100%;` only for zero states, this is working fine now cc @iamtarun1993

iamtarun1993: ok

iamtarun1993 (approved):

PR #492, opened 2021-07-20
FYLE-49b1rh: Replacing multi/user select cta buttons with header action buttons

Dimple16:
Screenshot 2021-07-20 at 9 06 22 PM
Screenshot 2021-07-20 at 9 51 23 PM

### DIFF
--- a/src/assets/styles/_fy_select.scss
+++ b/src/assets/styles/_fy_select.scss
@@ -80,8 +80,8 @@ $sub-title: #ababab;
 }
 
 %fy-select-cta {
-  --background: white;
-  --color: #ff3366;
+  --background: #{$pure-white};
+  --color: #{$primary};
   --box-shadow: none;
   text-transform: capitalize;
   margin-left: -16%;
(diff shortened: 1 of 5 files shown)


src/assets/styles/_fy_select.scss
```diff
@@ -78,3 +78,19 @@ $sub-title: #ababab;
   width: 94%;
   padding: 4px;
 }
+
+%fy-select-cta {
+  --background: white;
+  --color: #ff3366;
```
Dimple16, line 84: not using variables here as these are custom CSS properties of `ion-buttons` and not working with variables

Julias0: It works if you use interpolation
`--color: #{$primary}` or something.
I have used it in a few places

src/assets/styles/_fy_select.scss
```diff
@@ -78,3 +78,19 @@ $sub-title: #ababab;
   width: 94%;
   padding: 4px;
 }
+
+%fy-select-cta {
+  --background: white;
```
Julias0, line 83: use $pure-white

Julias0 (approved):

PR #502, opened 2021-07-26
FYLE-4bhwu9: Search not found zero state in select pop up

Dimple16:
**When the user has not selected any item**
Screenshot 2021-07-26 at 6 46 59 PM

**When the user has selected an item and wants to clear the selection**
- None option will be shown
Screenshot 2021-07-26 at 6 47 13 PM


src/app/shared/components/fy-select/fy-select-modal/fy-select-modal.component.html
```diff
@@ -27,6 +27,11 @@
 <ion-content>
   <div class="selection-modal--container">
 
+    <div *ngIf="(filteredOptions$|async)?.length === 0">
```
Julias0, line 30: <ng-container *ngIf="filteredOptions$|async as filteredOptions">
  <div *ngIf="filteredOptions.length === 0">
   // show zero state
  </div>
</ng-container>

Julias0 (approved):

PR #509, opened 2021-07-28
more lint fixes done

Julias0:


src/app/app.component.ts
```diff
@@ -61,6 +61,7 @@ export class AppComponent implements OnInit {
 
   previousUrl: string;
 
+  // eslint-disable-next-line max-params
```
Julias0, line 64: For constructor max params dosnt make sense. Fixed it for most of the other places exect 2-3 high impact ones. Have added todos there - will fix them off separately

abhishek1234321: we will need to keep adding this in every file? is there no other better way of doing this?

abhishek1234321 (approved):

PR #510, opened 2021-07-28
Add default payment mode

iamtarun1993:


src/app/fyle/capture-receipt/capture-receipt.page.ts
```diff
@@ -52,20 +53,46 @@ export class CaptureReceiptPage implements OnInit {
...
+          (orgSettings.advance_requests && orgSettings.advance_requests.enabled);
+
+      const userAccounts = this.accountsService.filterAccountsWithSufficientBalance(accounts, isAdvanceEnabled);
+      const isMultipleAdvanceEnabled = orgSettings && orgSettings.advance_account_settings &&
+      orgSettings.advance_account_settings.multiple_accounts;
+      const paymentModes = this.accountsService.constructPaymentModes(userAccounts, isMultipleAdvanceEnabled);
+
+      let account;
+
+      if (orgUserSettings?.preferences?.default_payment_mode === 'COMPANY_ACCOUNT') {
```
iamtarun1993, line 82: Added the reason in slack why `?`

abhishek1234321, line 82: `orgUserSettings?` when would this be `null`? didn't we add a type for this already?

src/app/fyle/capture-receipt/capture-receipt.page.ts
```diff
@@ -52,20 +53,46 @@ export class CaptureReceiptPage implements OnInit {
...
+        account = paymentModes.find(res => res.acc.displayName === 'Paid by Company');
+      } else if (orgUserSettings?.preferences?.default_payment_mode ==='PERSONAL_CORPORATE_CREDIT_CARD_ACCOUNT') {
```
abhishek1234321, line 84: what happens when this was set default earlier and then CCC was disabled?

abhishek1234321 (changes requested):

src/app/fyle/capture-receipt/capture-receipt.page.ts
iamtarun1993: `OrgUserSettings` model is there so `orgUserSettings?.` not neede

abhishek1234321: not understanding in what case would org_user_settings be not present?

abhishek1234321 (changes requested):

src/app/fyle/capture-receipt/capture-receipt.page.ts
iamtarun1993: Added isCCCEnabled ..
if CCC is disabled but the default mode is set to `PERSONAL_CORPORATE_CREDIT_CARD_ACCOUNT`
then it will set "Paid by me"

abhishek1234321 (approved):

PR #511, opened 2021-07-28
More and more lint fixes2

Julias0:


src/app/core/services/custom-inputs.service.ts
```diff
@@ -103,30 +102,81 @@ export class CustomInputsService {
...
-          displayValue = customProperty.value.display ? customProperty.value.display || '-' : '-';
-        } else {
-          displayValue = customProperty.value ? customProperty.value || '-' : '-';
-        }
-      }
-    } else if (customProperty.type === 'DATE') {
-      displayValue = customProperty ? this.datePipe.transform(customProperty.value, 'MMM dd, yyyy') : '-';
     }
 
+    if (customProperty.type === 'NUMBER') {
```
abhishek1234321, line 132: why remove `else if`? because every if will get executed and it really isn't needed

abhishek1234321 (approved):

PR #521, opened 2021-07-30
Sync expense fix

iamtarun1993:


src/app/fyle/capture-receipt/capture-receipt.page.ts
```diff
@@ -59,6 +60,12 @@ export class CaptureReceiptPage implements OnInit, OnDestroy {
 
   ngOnInit() {}
 
+  addMultipleExpensesToQueue(base64ImagesWithSource: Image[]) {
+    return from(base64ImagesWithSource).pipe(
+      mergeMap((res: Image) => this.addExpenseToQueue(res))
```
Julias0, line 65: use concatMap - this will cause parallal calls

Julias0: add a reduce after this otherwise any subscriber would execute multiple times

Julias0 (approved):

PR #523, opened 2021-08-02
FYLE-4fh1b5: Use expense fields for mileage distance unit instead of org settings

Dimple16:


abhishek1234321 (approved):

abhishek1234321 (approved):

src/app/fyle/add-edit-mileage/add-edit-mileage.page.ts
```diff
@@ -711,10 +711,11 @@ export class AddEditMileagePage implements OnInit {
       defaultVehicleType: defaultVehicle$,
       defaultMileageRate: defaultMileage$,
       currentEou: this.authService.getEou(),
-      autofillLocation: autofillLocation$
+      autofillLocation: autofillLocation$,
+      txnFields: this.txnFields$.pipe(take(1))
```
Julias0, line 715: I think this may lead to a cyclic dependency. Can we get on a call to resolve this

PR #525, opened 2021-08-03
FYLE-4fck3b CCC matched trasaction design changed

iravisankar:
localhost_8100_enterprise_add_edit_expense;id=txTX2wCI7srl;persist_filters=true(iPhone X) (2)


Julias0 (approved):

src/app/fyle/add-edit-expense/match-transaction/match-transaction.component.html
```diff
@@ -1,8 +1,8 @@
 <ion-header>
   <mat-icon class="fy-modal-top-bar" svgIcon="fy-rectangle"></mat-icon>
   <ion-toolbar class="fy-modal-toolbar">
```
Dimple16, line 3: we can use the 50% height modal here, let us use similar to payment mode

Dimple16 (approved):

PR #534, opened 2021-08-09
Bulk mode capture fix

iravisankar:


src/app/fyle/capture-receipt/capture-receipt.page.ts
```diff
@@ -221,7 +220,6 @@ export class CaptureReceiptPage implements OnInit, OnDestroy {
   }
 
   async review() {
-    await this.stopCamera();
```
iamtarun1993, line 224: We need to stop the camera when we are redirecting users to the review receipts page.
Else in iOS, there will be one green dot at the top navabr.

PR #558, opened 2021-08-18
Hide footer when keyboard is shown

yksolanki9:


yksolanki9:
Done by DImple in other pr

PR #565, opened 2021-08-19
Sentry appflow

iravisankar:


build_appflow.sh
```diff
@@ -0,0 +1,4 @@
+VERSION=`sentry-cli releases propose-version`
+ionic capacitor sync --prod --source-map
```
iamtarun1993, line 2: are you sure this was the command??

iamtarun1993: command was 
ionic capacitor build --prod --source-map
ionic capacitor sync

src/main.ts
```diff
@@ -11,7 +11,7 @@ import 'hammerjs';
 Sentry.init({
   dsn: environment.SENTRY_DSN,
   tracesSampleRate: 1.0,
-  release: 'please-replace-your-git-commit-version',
+  release: 'undefined',
```
iamtarun1993, line 14: why this become undefined ??

hooks/secrets.js
```diff
@@ -1,7 +1,8 @@
 const dotenv = require('dotenv');
 dotenv.config();
 const NATIVE_CONFIG = JSON.parse(process.env.NATIVE_CONFIG);
-
+console.log(process.env.CI_GIT_COMMIT_SHA);
```
iamtarun1993, line 4: do you need to print this?
or it is just for reference to see in app flow terminal so later we can debug if anything got broken.

Julias0 (approved):

Dimple16 (approved):

build_appflow.sh
iravisankar: `ionic capacitor build --prod --source-map ` we need mention android or ios for adding build ,
`ionic capacitor sync   --prod --source-map ` will take care of it for both android and ios build.

Julias0 (approved):

PR #573, opened 2021-08-23
Fix search bar and other minor changes

yksolanki9:
- Fixed case when the same expense is draft and has violated critical policy
- Fixed search bar
Before:
localhost_8100_enterprise_my_expenses_filters=%7B%7D(iPhone 6_7_8 Plus) (4)

localhost_8100_enterprise_my_expenses_filters=%7B%7D(iPhone 6_7_8 Plus) (3)

After:
localhost_8100_enterprise_my_dashboard(iPhone 6_7_8 Plus)

localhost_8100_enterprise_my_expenses_filters=%7B%7D(iPhone 6_7_8 Plus)

### DIFF
--- a/src/app/fyle/my-expenses/my-expenses.page.scss
+++ b/src/app/fyle/my-expenses/my-expenses.page.scss
@@ -89,10 +89,10 @@
 
   &--state-pill {
     width: 47%;
-    border: 1px solid #0077ee;
+    border: 1px solid $blue;
     padding: 8px;
     text-align: center;
-    color: #0077ee;
+    color: $blue;
     white-space: nowrap;
     &__left {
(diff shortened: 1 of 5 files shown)


src/app/fyle/my-expenses/my-expenses.page.scss
```diff
@@ -124,7 +126,7 @@
 
     &-icon {
       font-size: 18px;
-      margin-right: 8px;
+      margin: auto 8px auto 0;
```
Dimple16, line 129: this search bar alignment was breaking only in android, now that we are making generic changes, will need to test this change in both devices @yksolanki9

Dimple16 (approved):
Test this change in both the devices

PR #588, opened 2021-08-27
Added proper sources

Julias0:


src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -3372,6 +3246,15 @@ export class AddEditExpensePage implements OnInit {
           fileObj.type = (fileObj.type === 'application/pdf' || fileObj.type === 'pdf') ? 'pdf' : 'image';
           return fileObj;
         });
+
+        if (!(this.source.includes('_CAMERA') || this.source.includes('_FILE'))) {
```
abhishek1234321, line 3250: if an expense is created from web and edited from mobile would this overwrite things, it shouldn't IMHO

abhishek1234321 (approved):

PR #594, opened 2021-08-30
Update close icons, show clear all only when filter selected and remove * from boolean

yksolanki9:
Hide clear all button when no filter is selected:
https://user-images.githubusercontent.com/39493102/131331087-0818ca47-d539-4907-8c10-4d4d7ce41f42.mp4

Replace cross icons in headers with the new icons:
Before:
localhost_8100_enterprise_add_edit_expense;navigate_back=true(Moto G4) (1)

After:
localhost_8100_enterprise_add_edit_expense;navigate_back=true(Moto G4) (4)

Remove mandatory asterisk for booleans:
Before:
localhost_8100_enterprise_add_edit_expense;id=txzwYW2PNkYs;persist_filters=true(Moto G4) (4)

After:
localhost_8100_enterprise_add_edit_expense;id=txzwYW2PNkYs;persist_filters=true(Moto G4) (3)


src/assets/svg/fy-close.svg
```diff
@@ -1,4 +1,4 @@
 <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
Dimple16, line 1: remove height, width, and color, make sure it doesn't break in other places wherever used, verify everywhere

Dimple16 (changes requested):
few comments and resolve conflicts

Dimple16 (approved):

PR #639, opened 2021-09-13
minor tracking

pradeep8093:


src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -144,12 +146,36 @@ export class SwitchOrgPage implements OnInit, AfterViewInit, AfterViewChecked {
...
+        'User Org Name': originalEou.ou && originalEou.ou.org_name,
+        'User Org ID': originalEou.ou && originalEou.ou.org_id,
+        'User Full Name': originalEou.us && originalEou.us.full_name
+      };
+      this.trackingService.onSwitchOrg(properties);
+    });
+  };
+
   async switchToOrg(org: Org) {
+    const originalEou = await this.authService.getEou();
```
Julias0, line 170: put this inside the trackSwitchOrg method

Julias0 (approved):

src/app/auth/switch-org/switch-org.page.ts
pradeep8093: no this can't be put inside trackSwitchOrg method because the eou would change after switch org so kept outside

PR #649, opened 2021-09-15
Fix view comment header padding in ios

yksolanki9:
Before:
localhost_8100_enterprise_add_edit_expense;id=txisH8Ip9XQq;persist_filters=true(iPhone 6_7_8) (1)

After:
localhost_8100_enterprise_add_edit_expense;id=txisH8Ip9XQq;persist_filters=true(iPhone 6_7_8)


yksolanki9:
Done in other pr

PR #666, opened 2021-09-22
FYLE-cqkder: Fixed the spacing of loader and text in delete pop up

MasterYushi:
### Before: 
image

### After: 
image

### DIFF
--- a/src/global.scss
+++ b/src/global.scss
@@ -357,8 +357,7 @@ textarea:focus {
   height: 14px;
   display: inline-block;
   vertical-align: middle;
-  margin-left: 4px;
-  margin-right: 4px;
+  margin: 0 4px;
   animation: spin 2s linear infinite;
 }
 
--- a/src/global.scss
+++ b/src/global.scss
@@ -357,7 +357,7 @@ textarea:focus {
   height: 14px;
   display: inline-block;
   vertical-align: middle;
-  margin-left: 4px;
+  margin-inline: 4px;
   animation: spin 2s linear infinite;
 }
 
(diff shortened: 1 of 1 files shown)


src/global.scss
```diff
@@ -357,7 +357,7 @@ textarea:focus {
   height: 14px;
   display: inline-block;
   vertical-align: middle;
-  margin-left: 4px;
+  margin-inline: 4px;
```
abhishek1234321, line 360: the browser support for this isn't great https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline, could you check if what browser does ionic 5 / capacitor uses and check if this property is supported there

if not, we should stick with margin-left or right,

abhishek1234321 (changes requested):

src/global.scss
MasterYushi: ### I'm not sure. This is the only info I could find in the docs: 
image

### But I'll change it to `margin-left` and `margin-right` to be safe.

src/global.scss
```diff
@@ -358,6 +358,7 @@ textarea:focus {
   display: inline-block;
   vertical-align: middle;
   margin-left: 4px;
+  margin-right: 4px;
```
Dimple16, line 361: `margin: 0 4px;`

Dimple16: can remove these:
```
margin-left: 4px;
margin-right: 4px;
```

Dimple16: refer https://developer.mozilla.org/en-US/docs/Web/CSS/margin for margin shorthand properties @MasterYushi

Dimple16 (approved):

abhishek1234321 (approved):

PR #670, opened 2021-09-22
FYLE-cqkpw9: Fixed purpose field error alignment in per diem page

MasterYushi:
### Before
image

### After
image


src/global.scss
```diff
@@ -327,6 +327,10 @@ textarea:focus {
   margin: 0;
 }
 
+.mt-1 {
```
abhishek1234321, line 330: I would like us to avoid this, since we may not reuse this a lot

abhishek1234321 (changes requested):

src/global.scss
MasterYushi: The only other way would be to add inline styles. Do you want me to do that?

MasterYushi: @abhi The only other way would be to add inline styles. Do you want me to do that?

MasterYushi: > I would like us to avoid this, since we may not reuse this a lot

@abhishek1234321 The only other way would be to add inline styles. Do you want me to do that?

src/app/fyle/add-edit-per-diem/add-edit-per-diem.page.html
```diff
@@ -444,6 +454,7 @@
                         <div
                           *ngIf="fg.controls.purpose.touched && !fg.controls.purpose.valid"
                           class="add-edit-per-diem--error"
+                          style="margin-top: 4px"
```
Dimple16, line 457: @MasterYushi add a class like `add-edit-per-diem--purpose-error` in the scss class, we don't use inline-styles

Dimple16 (approved):

PR #674, opened 2021-09-22
Angular upgrade2

Julias0:


Julias0:
The Checks are failing due to a bug in the types of the library. I will look into this

PR #681, opened 2021-09-24
FYLE-cx5aqh: fixed the report name error alignment

MasterYushi:
image


Dimple16 (changes requested):
can we change the error font size to 12px and reduce top spacing a bit?

MasterYushi:
@Dimple16 Will reduce the font size, the top spacing is consistent with all other error messages in the app : 4px

Dimple16:
@MasterYushi error message top spacing should be `2px`

Julias0 (approved):

PR #760, opened 2021-11-22
refiner ROW support added

pradeep8093:


src/app/core/services/refiner.service.ts
```diff
@@ -39,15 +56,29 @@ export class RefinerService {
     this.isConnected$ = concat(that.networkService.isOnline(), networkWatcherEmitter.asObservable());
   }
 
+  getRegion(homeCurrency: string) {
+    if (homeCurrency === 'INR') {
+      return 'India';
+    } else if (this.americasCurrencyList.indexOf(homeCurrency) > -1) {
```
Julias0, line 62: use includes instead of indexOf

src/app/core/services/refiner.service.ts
```diff
@@ -39,15 +56,29 @@ export class RefinerService {
...
+      return 'International Americas';
+    } else if (this.europeCurrencyList.indexOf(homeCurrency) > -1) {
+      return 'Europe';
+    } else if (this.apacCurrencyList.indexOf(homeCurrency) > -1){
+      return 'International APAC';
+    } else {
```
Julias0, line 68: make this explicit

Julias0: give a region of unknown. That will help find and migrate buggy cases if we hit one

Julias0 (approved):

PR #761, opened 2021-11-22
Switch-org page: Add back btn, search animation, make search sticky, fix shimmers and other minor fixes 

yksolanki9:
https://www.loom.com/share/6e60e746a76c444e8551992cb362ccd7

### DIFF
--- a/src/app/auth/switch-org/switch-org.page.html
+++ b/src/app/auth/switch-org/switch-org.page.html
@@ -19,17 +19,18 @@
 </ion-header>
 
 <ion-content>
-  <ng-container *ngIf="!isLoading; else skeletonActiveOrg">
-    <ng-container *ngIf="activeOrg$ | async as activeOrg">
-      <ng-container *ngIf="primaryOrg$ | async as primaryOrg">
-        <app-active-org-card [org]="activeOrg" [isPrimaryOrg]="activeOrg.id === primaryOrg.id"></app-active-org-card>
+  <div #activeOrgCard>
+    <ng-container *ngIf="!isLoading; else skeletonActiveOrg">
+      <ng-container *ngIf="activeOrg$ | async as activeOrg">
+        <ng-container *ngIf="primaryOrg$ | async as primaryOrg">
+          <app-active-org-card [org]="activeOrg" [isPrimaryOrg]="activeOrg.id === primaryOrg.id"></app-active-org-card>
+        </ng-container>
       </ng-container>
     </ng-container>
-  </ng-container>
-
-  <ng-template #skeletonActiveOrg>
-    <app-active-org-card [isLoading]="true"></app-active-org-card>
-  </ng-template>
+    <ng-template #skeletonActiveOrg>
+      <app-active-org-card [isLoading]="true"></app-active-org-card>
+    </ng-template>
+  </div>
 
   <div class="switch-org__content-container">
     <div #content class="switch-org__content-container__block switch-org__content-container__content-block">
--- a/src/app/auth/switch-org/switch-org.page.html
+++ b/src/app/auth/switch-org/switch-org.page.html
@@ -1,9 +1,7 @@
 <ion-header mode="md">
   <ion-toolbar mode="md">
-    <ion-buttons slot="start">
-      <ion-button>
-        <ion-icon src="../../../assets/svg/arrow-back.svg"></ion-icon>
-      </ion-button>
+    <ion-buttons *ngIf="navigateBack" slot="start">
+      <ion-back-button mode="md"></ion-back-button>
     </ion-buttons>
     <div class="switch-org__toolbar__header-container">
       <ion-icon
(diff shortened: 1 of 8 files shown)


src/app/auth/switch-org/switch-org.page.html
```diff
@@ -103,13 +102,17 @@
...
     <ng-template #zeroState>
-      <app-fy-zero-state
-        header="Sorry, no results found!"
-        [message]="'We could not find what you were looking for.\nKindly check the keywords again.'"
-        image="../../../assets/images/zero-states/search-not-found.svg"
-        [unscaledImage]="true"
-        [useNewStyling]="true"
-      ></app-fy-zero-state>
+      <ng-container *ngIf="orgs$ | async as orgs">
+        <ng-container *ngIf="orgs.length > 1">
```
Dimple16, line 106: are we showing the zero state when `orgs.length > 1` ?

src/app/auth/switch-org/switch-org.page.scss
```diff
@@ -23,12 +23,44 @@
   }
 
   &__content-container {
-    padding: 16px;
+    position: sticky;
```
Dimple16, line 26: need to verify this on iOS

Dimple16 (changes requested):
few comments

src/app/auth/switch-org/switch-org.page.html
yksolanki9: Loom video: https://www.loom.com/share/c1a71588cced4975b15ff7c06890bcdd
- When a user logs in, we redirect them to the switch org page straight away. Now, if the user is part of only one org, we show a loader on top of the switch org page and user is logged into that org.
- Since the user is not part of any other org, `filteredOrgs` is an empty array and so a zero state will be shown
- This check hides the zero state if user is part of only one org.

src/app/auth/switch-org/switch-org.page.scss
yksolanki9: Will share ipa in a while. Thanks:)

Dimple16 (approved):

src/app/auth/switch-org/switch-org.page.html
```diff
@@ -34,11 +32,11 @@
...
 
   <div class="switch-org__content-container">
-    <div [hidden]="isSearchBarShown">
+    <div class="switch-org__content-container__block switch-org__content-container__content-block">
       <div class="switch-org__content-container__content d-flex">
-        <div>
+        <div class="switch-org__content-container__content__sub-block">
           <div class="switch-org__content-container__content__title">Switch Organization</div>
-          <div class="switch-org__content-container__content__sub-title">
+          <div class="text-ellipsis switch-org__content-container__content__sub-title">
```
abhishek1234321, line 39: why is this needed `text-ellipsis`?

src/app/auth/switch-org/switch-org.page.html
```diff
@@ -49,8 +47,7 @@
         ></ion-icon>
       </div>
     </div>
-
-    <div [hidden]="!isSearchBarShown">
+    <div #searchDiv class="switch-org__content-container__block switch-org__content-container__search-block">
```
abhishek1234321, line 50: #search is enough IMHO

src/app/auth/switch-org/switch-org.page.html
```diff
@@ -67,11 +64,11 @@
               [(ngModel)]="searchInput"
               matInput
               placeholder="Search for an org"
-              (focus)="onSearchBarFocus()"
               autocomplete="off"
             />
             <button
-              (click)="clearText('isFromSearchBar')"
+              (click)="clearSearchInput()"
```
abhishek1234321, line 70: clearSearch or resetSearch or reset?

abhishek1234321 (approved):

src/app/auth/switch-org/switch-org.page.html
yksolanki9: Done

src/app/auth/switch-org/switch-org.page.html
yksolanki9: Done

src/app/auth/switch-org/switch-org.page.html
yksolanki9: - The issue is that for smaller devices (Iphone 5), the text - `Select the organization you want to switch to` continues in the next line.
-  And since we are putting search bar on top, overflown text is visible (ss1 below).
- Since it works fine on other devices Moto G4 and above, I went with hiding the overflowing text with ellipsis(ss2 below)
- Other solution will be to change height of searchbar based on the height of text content.

localhost_8100_auth_sign_in(iPhone 5_SE) (1)

localhost_8100_auth_sign_in(iPhone 5_SE)

PR #764, opened 2021-11-23
Fyle 1v1hkjc : Fixed 'Please add a project ' alignment in Expenses

siddharth0-0:
image

image

### DIFF
--- a/src/app/fyle/add-edit-mileage/add-edit-mileage.page.html
+++ b/src/app/fyle/add-edit-mileage/add-edit-mileage.page.html
@@ -255,30 +255,28 @@
             </ng-container>
           </div>
 
-          <ng-container>
-            <div class="add-edit-mileage--project-container">
+          <ng-container *ngIf="txnFields$|async as txnFields">
+            <div class="add-edit-mileage--project-container" *ngIf="txnFields?.project_id">
               <ng-container *ngIf="isConnected$|async">
-                <ng-container *ngIf="txnFields$|async as txnFields">
-                  <ng-container
-                    *ngIf="txnFields?.project_id && (isProjectsEnabled$|async) && (((isIndividualProjectsEnabled$|async) && (individualProjectIds$|async)?.length > 0) || !(isIndividualProjectsEnabled$|async))"
-                  >
-                    <ng-container *ngIf="projectCategoryIds$|async as projectCategoryIds">
-                      <ng-container *ngIf="isProjectVisible$|async">
-                        <div
-                          class="add-edit-mileage--internal-block add-edit-mileage--project-block"
-                          [ngClass]="{'add-edit-mileage--no-project': !fg.controls.project.value}"
-                        >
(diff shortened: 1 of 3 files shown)


Dimple16:
Is this tested for the below cases:
- When projects are assigned 
- When projects are not assigned but it is mandatory 
- When billable is checked and not checked
- Project field width should change based on whether billable is present or not

Dimple16 (changes requested):
verify all the cases

siddharth0-0:
> Is this tested for the below cases:
> 
> * When projects are assigned  - Tested and able to save the expense
> * When projects are not assigned but it is mandatory - Tested and message is hidden in this case
> * When billable is checked and not checked - Tested and able to save expense
> * Project field width should change based on whether billable is present or not - Tested and looks fine

Dimple16 (approved):

src/app/fyle/add-edit-mileage/add-edit-mileage.page.html
```diff
@@ -255,17 +255,17 @@
...
                 <ng-container
-                  *ngIf="txnFields?.project_id && (isProjectsEnabled$|async) && (((isIndividualProjectsEnabled$|async) && (individualProjectIds$|async)?.length > 0) || !(isIndividualProjectsEnabled$|async))"
+                  *ngIf="(isProjectsEnabled$|async) && (((isIndividualProjectsEnabled$|async) && (individualProjectIds$|async)?.length > 0) || !(isIndividualProjectsEnabled$|async))"
                 >
                   <ng-container *ngIf="projectCategoryIds$|async as projectCategoryIds">
                     <ng-container *ngIf="isProjectVisible$|async">
                       <div
                         class="add-edit-mileage--internal-block add-edit-mileage--project-block"
-                        [ngClass]="{'add-edit-mileage--no-project': fg.controls.project.touched && !fg.controls.project.valid}"
+                        [ngClass]="{'add-edit-mileage--no-project': !fg.controls.project.value}"
```
Julias0, line 268: why change this condition?

siddharth0-0: In Css  --no-project is set to 'margin-right: 0px !important' 
Earlier when was selecting project as 'None' it was giving a certain margin which was making select dropdown look short 

InkedScreenshot_20211129-115502_Fyle_LI

Julias0: but this css class will still be added under certain conditions right? Will it look ok in those cases?

siddharth0-0: Yes, --no-project will only be used under when the project value is none.
This is the only time when this --no-project class will be used, In other condition when project value is !Null there will be an checkbox option on right.
image
In this condition the select dropdown will have 'margin-right: 20px;'

Julias0 (approved):

PR #779, opened 2021-12-08
FYLE-1uvpzma: Remove add more using pop up in iOS and use device pop up to add receipts

Dimple16:


Dimple16:
**Before**
image_123986672
image_123986672 (1)

**After**
image0 (12)
image1 (7)

Julias0 (approved):

PR #793, opened 2021-12-21
Make match and add_to_report calls serial

modernwarfareuplink:


modernwarfareuplink:
Added here https://github.com/fylein/fyle-mobile-app2/pull/794

PR #794, opened 2021-12-21
Make match and add_to_report calls serial

modernwarfareuplink:


src/app/core/services/transactions-outbox.service.ts
```diff
@@ -325,6 +325,21 @@ export class TransactionsOutboxService {
     });
   }
 
+  matchIfRequired(transactionId, entry) {
```
Dimple16, line 328: add argument types

Dimple16 (approved):

src/app/core/services/transactions-outbox.service.ts
```diff
@@ -325,6 +325,21 @@ export class TransactionsOutboxService {
     });
   }
 
+  matchIfRequired(transactionId: string, cccId: string) {
```
Julias0, line 328: lets avoid promises since we want to heavily use rxjs pipelines
```typescript
return iif(() => cccId, this.transactionService.matchCCCExpense(transactionId, cccId), of(null));
```

Julias0 (approved):

PR #801, opened 2021-12-23
Add loader for receipt and reduce loader size

yksolanki9:
In case of slow connections the receipts inside edit expense take too long to load, so we needed a way to tell the user that the receipt is loading. There are two parts of the solution:
-  The receipt preview box was taking too long to load. This was because `attachments$` which depends on `loadAttachments$` was emitting a value after a timeout. So I changed loadAttachments to a BehaviourSubject and so it emits the first value on declaration.
- To show a loader while the receipt is being loaded, I am using the image `load` event. So while the first image is loading, we show a `Loading Receipt` spinner. And all the subsequent receipts have a background loading gif which will he hidden as and when the receipt loads. After all receipts are loaded the background gif is removed.

https://www.loom.com/share/2e6a14bf32ad46308ae9167ed875f084


src/app/shared/components/receipt-preview-thumbnail/receipt-preview-thumbnail.component.html
```diff
@@ -8,11 +8,16 @@
   </div>
 </div>
 
-<div class="receipt-preview-thumbnail--slider-container">
-  <div class="receipt-preview-thumbnail--spinner-overlay" *ngIf="isUploading">
+<div
+  class="receipt-preview-thumbnail--slider-container"
+  [ngClass]="{ 'receipt-loader': numLoadedImage === attachments.length ? false : true }"
```
abhishek1234321, line 13: `numLoadedImage !== attachments.length` should do no?

abhishek1234321 (approved):

src/app/shared/components/receipt-preview-thumbnail/receipt-preview-thumbnail.component.html
yksolanki9: Not sure why but it wont work

iravisankar (approved):

src/app/shared/components/receipt-preview-thumbnail/receipt-preview-thumbnail.component.html
```diff
@@ -46,7 +51,7 @@
           disableZoomControl="disable"
           backgroundColor="white"
         >
-          <img class="receipt-preview-thumbnail--image" [src]="attachment?.url" />
+          <img (load)="onLoad()" class="receipt-preview-thumbnail--image" [src]="attachment?.url" />
```
Julias0, line 54: This method of showing a loader till onLoad is called - this looks a bit funky. 
Why are we doing it like this?

src/app/shared/components/receipt-preview-thumbnail/receipt-preview-thumbnail.component.ts
```diff
@@ -28,6 +28,8 @@ export class ReceiptPreviewThumbnailComponent implements OnInit, DoCheck {
 
   previousCount: number;
 
+  numLoadedImage = 0;
```
Julias0, line 31: set the value inside ngOnInit please

Julias0 (approved):

PR #802, opened 2021-12-26
fixed : Send Report Id From View Report on click of add expense

siddharth0-0:
No Visual Changes

### DIFF
--- a/src/app/fyle/add-edit-expense/add-edit-expense.page.ts
+++ b/src/app/fyle/add-edit-expense/add-edit-expense.page.ts
@@ -1286,17 +1286,25 @@ export class AddEditExpensePage implements OnInit {
       )
     );
     const selectedReport$ = this.etxn$.pipe(
-      switchMap((etxn) =>
-        iif(
-          () => etxn.tx.report_id,
-          this.reports$.pipe(
+      switchMap((etxn) => {
+        if (etxn.tx.report_id) {
+          return this.reports$.pipe(
             map((reportOptions) =>
               reportOptions.map((res) => res.value).find((reportOption) => reportOption.rp.id === etxn.tx.report_id)
             )
-          ),
-          of(null)
-        )
-      )
+          );
+        } else if (!etxn.tx.report_id && this.activatedRoute.snapshot.params.rp_id) {
+          return this.reports$.pipe(
+            map((reportOptions) =>
+              reportOptions
+                .map((res) => res.value)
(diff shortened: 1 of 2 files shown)


Dimple16 (approved):

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -1296,6 +1296,19 @@ export class AddEditExpensePage implements OnInit {
           ),
           of(null)
         )
+      ),
+      switchMap(() =>
+        iif(
+          () => this.activatedRoute.snapshot.params.rp_id,
```
Julias0, line 1302: wait - the reportOption which was selected at the above level, that is not getting considered at all. That will have some implications no? What if the report id has not been passed in the URL? This looks incorrect to me

Dimple16 (changes requested):
check the cases discussed

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -1449,7 +1457,7 @@ export class AddEditExpensePage implements OnInit {
             paymentMode: selectedPaymentMode$,
             project: selectedProject$,
             category: selectedCategory$,
-            report: selectedReport$,
+            report: this.activatedRoute.snapshot.params.rp_id ? paramReport$ : selectedReport$,
```
Dimple16, line 1460: can you check if there is any case where the `params.rp_id` is set when this page opens? because according to the condition priority is given to the param report and not the selected report. Ideally, this check should happen in `selectedreport$` 
Few things to check:
- If it is a new expense if there is no existing selected report if the report param is present -> set the report field
- Once the page is open, it'll have the report field filled, now on edit, it should be added to any other report if I make a change - because there might be many reports in the list
- Check cases of adding expense and edit expense - if the correct report is filled/modified correctly when edited
- While opening an expense (edit expense page) -> the correct report should be selected

Dimple16 (changes requested):
verify the conditions mentioned

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
siddharth0-0: Tested all the test cases

Dimple16 (approved):

abhishek1234321 (approved):

PR #829, opened 2022-01-20
FYLE-1z55me7: Prompt report submission mobile app changes

Dimple16:
### List of changes

- Making `Add to report` mandatory during add expense
- Adding a different background to the `Add to report` section in the expense form
- On Save and new, redirect to expense form with report name filled

Screenshot 2022-01-20 at 7 50 15 PM

Screenshot 2022-01-20 at 7 53 28 PM

### DIFF
--- a/src/app/fyle/add-edit-expense/add-edit-expense.page.html
+++ b/src/app/fyle/add-edit-expense/add-edit-expense.page.html
@@ -956,12 +956,19 @@
           <ng-container *ngIf="!(isCriticalPolicyViolated$|async)">
             <ng-container>
               <ng-container *ngIf="reports$|async as reports">
-                <div class="add-edit-expense--internal-block">
+                <div class="add-edit-expense--internal-block add-edit-expense--report-section">
                   <app-fy-add-to-report
                     [label]="etxn.tx.report_id ? 'Report' : 'Add to Report'"
                     [options]="reports"
                     formControlName="report"
+                    [mandatory]="isReportMandatory"
                   ></app-fy-add-to-report>
+                  <div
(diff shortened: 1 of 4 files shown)


Julias0 (approved):

src/app/fyle/add-edit-expense/add-edit-expense.page.html
```diff
@@ -956,12 +956,19 @@
           <ng-container *ngIf="!(isCriticalPolicyViolated$|async)">
             <ng-container>
               <ng-container *ngIf="reports$|async as reports">
-                <div class="add-edit-expense--internal-block">
+                <div class="add-edit-expense--internal-block add-edit-expense--report-section">
```
Julias0, line 959: What happens if an expense 
hits critical policy violation during adding of a new expense
No reports are present
No internet is present
Wont it lead to the case that this field is a required field but it is not visible?

Julias0 (changes requested):
Put out a note in #mobile_app for some questions

src/app/fyle/add-edit-expense/add-edit-expense.page.html
Dimple16: The cases are handled:
- When no reports are present
   - User can create a new report
-  No internet is present
   - Add to Report field is made mandatory only when online
-  When an expense is critical policy violated
   -  If a report is added to expense, on click of the save button, the critical policy violation pop up is shown
   -  On continuing to save with the critical policy violation, expense is saved and not added to report

Julias0 (approved):

PR #851, opened 2022-01-31
update project name in search project modal

Ajinkya2000:
localhost_8100_enterprise_add_edit_expense;id=tx96JLRvfXqC;persist_filters=true(iPhone SE) (1)


src/app/shared/components/fy-select-project/fy-select-project.component.ts
```diff
@@ -33,6 +33,8 @@ export class FySelectProjectComponent implements OnInit, ControlValueAccessor, O
 
   @Input() recentlyUsed: { label: string; value: ExtendedProject; selected?: boolean }[];
 
+  @Input() placeholder = 'Select Project';
```
pradeep8093, line 36: why hardcoding project

Ajinkya2000: Well placeholder is being fetched from the expenseField. So just to make sure if placeholder is not provided in props, it always has a default value.

shishirams222 (approved):

PR #853, opened 2022-01-31
Update Project Name in Split Expense

Ajinkya2000:
localhost_8100_enterprise_add_edit_expense;id=tx96JLRvfXqC;persist_filters=true(iPhone SE)
localhost_8100_enterprise_add_edit_expense;id=tx96JLRvfXqC;persist_filters=true(iPhone SE) (1)

### DIFF
--- a/src/app/fyle/split-expense/split-expense.page.html
+++ b/src/app/fyle/split-expense/split-expense.page.html
@@ -125,7 +125,8 @@
               *ngIf="splitExpenseForm.controls.project.touched && !splitExpenseForm.controls.project.valid"
               class="split-expense--error"
             >
-              Please select {{ selectProjectMessage }}.
+              Please select {{ this.txnFields?.project_id?.field_name === 'Project' ? 'a project' :
+              this.txnFields?.project_id?.field_name }}.
             </div>
           </div>
 
--- a/src/app/fyle/split-expense/split-expense.page.html
+++ b/src/app/fyle/split-expense/split-expense.page.html
@@ -8,7 +8,9 @@
       </ion-buttons>
 
       <div class="split-expense--header-block">
-        <div class="split-expense--header-title" mode="md">Split by {{ splitType }}</div>
+        <div class="split-expense--header-title" mode="md">
+          Split by {{ splitType === 'projects' ? txnFields.project_id.field_name : splitType}}
+        </div>
         <ion-buttons>
           <ion-button
             (click)="save()"
@@ -112,7 +114,8 @@
 
           <div class="split-expense--internal-block" *ngIf="splitType === 'projects'">
             <app-fy-select-project
-              [label]="'Project'"
+              [label]="txnFields?.project_id?.field_name"
+              [placeholder]="txnFields?.project_id?.placeholder"
               formControlName="project"
               [mandatory]="true"
               [cacheName]="'splitExpenseProjectCache'"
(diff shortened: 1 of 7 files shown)


pradeep8093 (approved):

src/app/fyle/split-expense/split-expense.page.ts
```diff
@@ -71,6 +71,8 @@ export class SplitExpensePage implements OnInit {
 
   completeTxnIds: string[];
 
+  selectProjectMessage = 'a Project';
```
shishirams222, line 74: This can be renamed as something like: `selectProjectMessage` -> `warningTextLabel`

Ajinkya2000: Okay will change this.

src/app/fyle/split-expense/split-expense.page.ts
```diff
@@ -454,6 +456,11 @@ export class SplitExpensePage implements OnInit {
         this.setValuesForCCC(currencyObj, homeCurrency, isCorporateCardsEnabled);
       });
     });
+
+    // if the project field name is not project, this will remove "a" from the required text
```
Dimple16, line 460: why not have the `a` for both cases? is removing required here?

src/app/fyle/split-expense/split-expense.page.html
```diff
@@ -122,7 +125,7 @@
               *ngIf="splitExpenseForm.controls.project.touched && !splitExpenseForm.controls.project.valid"
               class="split-expense--error"
             >
-              Please select a project.
+              Please select {{ selectProjectMessage }}.
```
Dimple16, line 128: is this required here? can we not directly use it here? you can add it conditionally it

src/app/fyle/split-expense/split-expense.page.html
```diff
@@ -112,7 +114,8 @@
 
           <div class="split-expense--internal-block" *ngIf="splitType === 'projects'">
             <app-fy-select-project
-              [label]="'Project'"
+              [label]="txnFields?.project_id?.field_name"
+              [placeholder]="txnFields?.project_id?.placeholder"
```
Dimple16, line 118: `placeholder` is not required here, only `label` should do, the component handles what has to be shown as the placeholder

Dimple16: check `fy-select-project.component.html`

src/app/fyle/split-expense/split-expense.page.html
```diff
@@ -8,7 +8,9 @@
       </ion-buttons>
 
       <div class="split-expense--header-block">
-        <div class="split-expense--header-title" mode="md">Split by {{ splitType }}</div>
+        <div class="split-expense--header-title" mode="md">
+          Split by {{ splitType === 'projects' ? txnFields.project_id.field_name : splitType}}
```
Dimple16, line 12: `txnFields?.project_id?.field_name`

Dimple16 (changes requested):

src/app/fyle/split-expense/split-expense.page.ts
Ajinkya2000: Yes, because if a user renamed `project` to say `event`, then ideally we must be showing **an event** not **a event**. So we decided to just remove the `a`.

src/app/fyle/split-expense/split-expense.page.html
Ajinkya2000: Yes added conditionally. It made more sense.

src/app/fyle/split-expense/split-expense.page.html
Ajinkya2000: So placeholder is required here because the admin also sets the placeholder text. So we need to show that.

Dimple16 (approved):

PR #867, opened 2022-02-02
deprecated DE endpoint

madhavmansuriya40:


src/app/core/services/transactions-outbox.service.ts
```diff
@@ -491,30 +491,16 @@ export class TransactionsOutboxService {
...
+    const url = this.ROOT_ENDPOINT + '/data_extractor/extract';
     let suggestedCurrency = null;
     const fileName = fileType && fileType === 'pdf' ? '000.pdf' : '000.jpeg';
 
     // send homeCurrency of the user's org as suggestedCurrency for data-extraction
     return this.offlineService
       .getHomeCurrency()
       .toPromise()
-      .then(({ homeCurrency }) =>
-        this.offlineService
```
madhavmansuriya40, line 503: we don't need org settings now, as we are not checking for any DE service enabled
we care deprecating old DE service so just making the code same and updating endpoint

Dimple16 (approved):

Julias0 (approved):

PR #878, opened 2022-02-07
Disable New trip request cta and update deprecation msg

yksolanki9:
For orgs without extension:
localhost_8100_enterprise_my_trips(Pixel 5)

For orgs with extension:
localhost_8100_enterprise_my_trips(Pixel 5) (1)

### DIFF
--- a/src/app/core/services/trip-requests.service.ts
+++ b/src/app/core/services/trip-requests.service.ts
@@ -499,7 +499,7 @@ export class TripRequestsService {
     return this.apiService.get('/trip_requests', data);
   }
 
-  getTripDeprecationMsg(view: 'individual' | 'team') {
+  doesOrgHaveExtension() {
     const orgsWithExtension = [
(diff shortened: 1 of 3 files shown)


src/app/core/services/trip-requests.service.ts
```diff
@@ -519,21 +519,26 @@ export class TripRequestsService {
       'orAQ0g05flq6',
       'orcnfOH4mR5t',
     ];
+    return from(this.authService.getEou()).pipe(map((eou) => orgsWithExtension.indexOf(eou.ou.org_id) > -1));
```
Julias0, line 522: `orgsWithExtension.indexOf(eou.ou.org_id) > -1` use `includes`

Julias0 (approved):

src/app/core/services/trip-requests.service.ts
yksolanki9: Done

PR #896, opened 2022-02-14
Fix: Wrong amount calculation when user creates expense in past

KirtiGautam:


alamirfan (approved):

alamirfan (approved):

viiicky (approved):

viiicky (approved):

src/app/core/services/transaction.service.ts
```diff
@@ -496,6 +497,13 @@ export class TransactionService {
         if (etxn.tx_tax) {
           delete etxn.tx_tax;
         }
+
+        if (etxn.tx_created_at) {
+          // Just stripping off timezone since created_at is
+          // malformed(time zone offset and time stamp pair is incorrect).
+          etxn.tx_created_at = moment(etxn.tx_created_at).format('YYYY-MM-DDTHH:mm:ss.SSS') + 'Z'; // Z token is reserved for offset
```
Julias0, line 504: This is very hacky. The right way to do this is to use the .utc method inside moment and set hours, mins and secs

viiicky: hey no; we need to do this hack; because the frontend created malformed dates. This the way to do it, unless we fix of the base implementation - which is a gigantic change (the whole UI code is written on top of that wrong implementation)

Julias0 (approved):

KirtiGautam:
Found a major bug while testing, won't be moving forward with this approach. Hence, closing this off.

PR #897, opened 2022-02-14
Added text-overflow-ellipsis for report name in the list of reports in the add to report dialog

saitejasFyle:


saitejasFyle:
Before:
Screenshot_20220210-134933

After:
Screenshot 2022-02-16 at 11 28 50 AM

Dimple16 (approved):
need to check this on smaller dimension devices, can we verify this and only then merge?

saitejasFyle:
> need to check this on smaller dimension devices, can we verify this and only then merge?

I've checked it with all devices in the responsvie section in the browser.

PR #903, opened 2022-02-15
Fix alignment of Please enter amount text

yksolanki9:
Before:
localhost_8100_enterprise_add_edit_advance_request(Pixel 5)

After:
localhost_8100_enterprise_add_edit_mileage;navigate_back=true(Pixel 5)

### DIFF
--- a/src/app/fyle/add-edit-advance-request/add-edit-advance-request.page.scss
+++ b/src/app/fyle/add-edit-advance-request/add-edit-advance-request.page.scss
@@ -183,7 +183,7 @@
   &--amount-error {
     color: $red;
     font-size: 12px;
-    margin: -8px 0 0 44%;
+    margin: -8px 0 0 52%;
   }
 
   &--primary-cta {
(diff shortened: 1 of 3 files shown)


src/app/fyle/add-edit-advance-request/add-edit-advance-request.page.scss
```diff
@@ -183,7 +183,7 @@
   &--amount-error {
     color: $red;
     font-size: 12px;
-    margin: -8px 0 0 44%;
+    margin: -8px 0 0 52%;
```
Dimple16, line 186: need to verify this on different resolutions, might break in smaller width phones

Dimple16 (changes requested):

Dimple16 (approved):

src/app/fyle/add-edit-advance-request/add-edit-advance-request.page.scss
yksolanki9: Discussed

PR #917, opened 2022-02-22
currencies of an expense is given precedence over an org's default home currency

saitejasFyle:


Dimple16:
can you please add screenshots?

Dimple16 (approved):

Julias0 (approved):

PR #945, opened 2022-03-09
Fix `Assert: cannot be animating` error in mobile app

yksolanki9:
The error is because we are switching the swipe gesture while the menu controller is closing(animating). So adding a timeout of 500ms so that the swipe gesture is enabled/disabled after the animation

Ps - This also fixes the annoying debuggers we see on the browser while switching routes

Ref - https://github.com/ionic-team/ionic-framework/issues/19000
image


Julias0 (approved):

ekoeryanto:
ionic 7 raise this error again

PR #973, opened 2022-03-25
changed colour to white when connection is online

sandy-fyle:
when the app is offline the colour will back but when it is online the colour I set to white

before

after

### DIFF
--- a/src/app/shared/components/fy-connection/fy-connection.component.scss
+++ b/src/app/shared/components/fy-connection/fy-connection.component.scss
@@ -23,6 +23,6 @@
 
   &--online-message {
     background-color: $green;
-    color: #fff;
+    color: $pure-white;
   }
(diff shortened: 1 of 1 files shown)


src/app/shared/components/fy-connection/fy-connection.component.scss
```diff
@@ -23,5 +23,6 @@
 
   &--online-message {
     background-color: $green;
+    color: #fff;
```
Dimple16, line 26: use color variables `$pure-white`

Dimple16: add before and after screenshots

Dimple16 (changes requested):

Dimple16 (approved):

PR #977, opened 2022-03-28
Disable Insecure HTTP loads on localhost

yksolanki9:
Fix for issue - https://www.notion.so/fyleuniverse/Vapt-q1-2022-fa65f76bc48f48f68e2974c07f7738da#716c6413682b4ce58a4ffdd5222e1ae9

🍏 Tested on device

### DIFF
--- a/hooks/secrets.js
+++ b/hooks/secrets.js
@@ -138,14 +138,6 @@ module.exports = {
     <dict>
         <key>NSAllowsArbitraryLoads</key>
         <false/>
-        <key>NSExceptionDomains</key>
-        <dict>
-            <key>localhost</key>
-            <dict>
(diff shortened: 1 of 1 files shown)


hooks/secrets.js
```diff
@@ -138,14 +138,6 @@ module.exports = {
     <dict>
         <key>NSAllowsArbitraryLoads</key>
         <false/>
-        <key>NSExceptionDomains</key>
```
Dimple16, line 141: we should be removing this in `Info.plist` file too

Dimple16 (changes requested):

hooks/secrets.js
yksolanki9: Info.plist is generated at build time and it is not tracked by git

PR #978, opened 2022-03-28
Unify ccce expenses flow - Fixes-1

Dimple16:
Screenshot 2022-03-28 at 2 45 54 PM
Screenshot 2022-03-28 at 2 46 30 PM
Screenshot 2022-03-28 at 2 46 56 PM

### DIFF
--- a/src/app/fyle/add-edit-expense/add-edit-expense.page.html
+++ b/src/app/fyle/add-edit-expense/add-edit-expense.page.html
@@ -108,7 +108,9 @@
             <ion-row>
               <ion-col>
                 <div class="add-edit-expense--card-container__label">Card number</div>
-                <div class="add-edit-expense--card-container__value">{{ cardNumber | maskNumber }}</div>
+                <div class="add-edit-expense--card-container__value">
+                  {{ cardNumber ? (cardNumber | maskNumber) : '-' }}
+                </div>
               </ion-col>
               <ion-col>
                 <div class="add-edit-expense--card-container__label">Spend Date</div>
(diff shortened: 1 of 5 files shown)


src/app/fyle/add-edit-expense/add-edit-expense.page.html
```diff
@@ -1025,7 +1035,7 @@
 
       <ng-container *ngIf="etxn$|async as etxn">
         <ng-container
-          *ngIf="(etxn && etxn.tx.creator_id === 'SYSTEM_CORPORATE_CARD' && selectedCCCTransaction) && !(isUnifyCcceExpensesSettingsEnabled && isCccExpense)"
+          *ngIf="(etxn && etxn.tx.creator_id === 'SYSTEM_CORPORATE_CARD' && selectedCCCTransaction) && !isUnifyCcceExpensesSettingsEnabled"
```
saitejasFyle, line 1038: Why isCCCExpense has been removed?

src/app/fyle/view-expense/view-expense.page.html
```diff
@@ -65,7 +65,7 @@
         <div *ngIf="isProjectShown" class="view-expense--card-container__project">
           {{ projectFieldName | titlecase }}: {{ etxn.tx_project_name ? etxn.tx_project_name : 'Unspecified' }}
         </div>
-        <div *ngIf="isUnifyCcceExpensesSettingsEnabled && isCCCTransaction">
+        <div *ngIf="isUnifyCcceExpensesSettingsEnabled && isCCCTransaction && cardNumber">
```
saitejasFyle, line 68: What if card number is not present - Don't we have to display anything?

src/app/fyle/add-edit-expense/add-edit-expense.page.html
Dimple16: `selectedCCCTransaction` serves the same purpose, so removed the redundant check

src/app/fyle/view-expense/view-expense.page.html
Dimple16: this we need confirmation from design, as of now, showing only the label looked odd, so we'll be showing this section only if it's present

PR #991, opened 2022-03-29
Use secure storage plugin in capacitor

yksolanki9:


src/app/core/interceptors/httpInterceptor.ts
```diff
@@ -68,6 +70,7 @@ export class HttpConfigInterceptor implements HttpInterceptor {
       concatMap((refreshToken) => this.routerAuthService.fetchAccessToken(refreshToken)),
       catchError((error) => {
         this.userEventService.logout();
+        this.secureStorageService.clearAll();
         this.storageService.clearAll();
```
Dimple16, line 74: will we still need `storageService` here?

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -213,6 +215,7 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
           from(this.proceed()).subscribe(noop);
         },
         async (err) => {
+          await this.secureStorageService.clearAll();
           await this.storageService.clearAll();
```
Dimple16, line 219: aren't we replacing it?

src/app/core/services/secure-storage.service.ts
```diff
@@ -0,0 +1,45 @@
...
+      value: JSON.stringify(value),
+    });
+  }
+
+  async get(key: string) {
+    try {
+      const stringifiedObject = await SecureStoragePlugin.get({
+        key,
+      });
+      if (stringifiedObject && stringifiedObject.value) {
```
Dimple16, line 26: `stringifiedObject?.value`

Dimple16 (approved):

src/app/core/interceptors/httpInterceptor.ts
yksolanki9: Yes, we clear all locally stored data when logging out

src/app/auth/switch-org/switch-org.page.ts
yksolanki9: We clear all cached data on logout and when switching orgs

src/app/core/services/secure-storage.service.ts
yksolanki9: Done

Julias0 (approved):

PR #992, opened 2022-03-29
added auto-population of report name in create-expense-view

ayush-seth:
https://user-images.githubusercontent.com/24858182/160634788-96ba21a0-7d88-4e85-8ab2-ac4448bda7d2.mp4


src/app/shared/components/fy-add-to-report/fy-add-to-report.component.ts
```diff
@@ -127,11 +127,14 @@ export class FyAddToReportComponent implements OnInit, OnDestroy {
     }
 
     if (data && data.createDraftReport) {
+      const reportTitle = await this.reportService.getReportPurpose({ ids: null }).toPromise();
```
Dimple16, line 130: why are we converting observable to promise here? let us use it as an observable directly

Dimple16 (changes requested):

src/app/shared/components/fy-add-to-report/fy-add-to-report.component.ts
MasterYushi: because we need to wait for the api call to fetch the report name before displaying the create report modal. This is inside an async function so I am following the async await syntax

Dimple16: we can have an observable with the value of API response, on subscribe of that you can set the reportTitle, 
The change here is valid, but since we are following the observable pattern everywhere, I guess we can avoid converting it to promises again

Dimple16 (changes requested):

src/app/shared/components/fy-add-to-report/fy-add-to-report.component.ts
ayush-seth: We cannot do that. Subscribing to it and then setting the value of the variable to the returned value will not halt the execution.

If the network is slow, the modal will be called with 'reportTitle' as undefined before the API call is resolved that's why we need to use await here

Dimple16 (approved):

Julias0 (approved):

PR #1004, opened 2022-04-02
changed the font colour to white when offline

sandy-fyle:
changed the font colour to white when offline

Before:
before

After:
after

### DIFF
--- a/src/app/fyle/add-edit-expense/add-edit-expense.page.scss
+++ b/src/app/fyle/add-edit-expense/add-edit-expense.page.scss
@@ -271,7 +271,7 @@
   &--project-block {
     flex-grow: 1;
     margin-right: 26px;
-    margin-top: -13px;
+    margin-top: -12px;
   }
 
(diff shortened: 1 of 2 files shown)


src/app/fyle/add-edit-expense/add-edit-expense.page.scss
```diff
@@ -271,6 +271,7 @@
   &--project-block {
     flex-grow: 1;
     margin-right: 26px;
+    margin-top: -12px;
```
Dimple16, line 274: this shouldn't be here

Dimple16 (changes requested):

src/app/fyle/add-edit-expense/add-edit-expense.page.scss
sandy-fyle: oh! I created this branch from previous bug branch instead of master so it contains those changes as well. I will change that

PR #1016, opened 2022-04-06
use v2 spender_employees

Chethan-Fyle:


src/app/core/models/spender-employee.model.ts
```diff
@@ -0,0 +1,9 @@
+export interface SpenderEmployee {
```
Julias0, line 1: rename to Employee and move inside a folder called spender inside models folder

Julias0 (approved):

PR #1017, opened 2022-04-06
Fixed missing NPS survey triggers in Submit Report Flow.

OmkarJ13:


src/app/fyle/my-view-report/my-view-report.page.ts
```diff
@@ -391,7 +391,7 @@ export class MyViewReportPage implements OnInit {
     await popover.present();
 
     const { data } = await popover.onWillDismiss();
-    this.refinerService.startSurvey({ actionName: 'Resubmit Report' });
```
abhishek1234321, line 394: is this not needed here? or was it in the wrong place?

abhishek1234321 (approved):

src/app/fyle/my-view-report/my-view-report.page.ts
OmkarJ13: Yes, it was in the wrong place. Basically, the survey would show up even after canceling the popover and not resubmitting the report.

Dimple16 (approved):

PR #1036, opened 2022-04-14
fixed payment processing chip on advance report page

sandy-fyle:
fixed payment processing chip on advance report page
before
after


Dimple16:
this isn't `advance report page` no? do you mean team reports?

Dimple16 (approved):

PR #1043, opened 2022-04-18
Fix sentry error: `TypeError: Cannot read properties of undefined (reading 'rp_purpose')`

yksolanki9:
Could not repro the error, but this pr should fix it

### DIFF
--- a/src/app/shared/components/fy-view-report-info/fy-view-report-info.component.ts
+++ b/src/app/shared/components/fy-view-report-info/fy-view-report-info.component.ts
@@ -3,6 +3,7 @@ import { ModalController, Platform } from '@ionic/angular';
 import { ExtendedReport } from 'src/app/core/models/report.model';
 import { Expense } from 'src/app/core/models/expense.model';
 import { Observable, combineLatest } from 'rxjs';
+import { filter } from 'rxjs/operators';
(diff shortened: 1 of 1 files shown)


src/app/shared/components/fy-view-report-info/fy-view-report-info.component.ts
```diff
@@ -76,7 +77,7 @@ export class FyViewReportInfoComponent implements OnInit {
   ngOnInit(): void {}
 
   ionViewWillEnter() {
-    this.erpt$.subscribe((erpt) => {
+    this.erpt$.pipe(filter((erpt) => !!erpt)).subscribe((erpt) => {
```
Dimple16, line 80: should we add an `if` condition to check if erpt exists? OR use `erpt?.rp_purpose`

Dimple16 (changes requested):

src/app/shared/components/fy-view-report-info/fy-view-report-info.component.ts
yksolanki9: If erpt is does not exist, the filter condition wont be satisfied and no value will be emitted

Dimple16 (approved):

PR #1055, opened 2022-04-25
Capacitor 3 upgrade

yksolanki9:


src/app/core/services/device.service.ts
```diff
@@ -12,6 +12,21 @@ export class DeviceService {
...
 
   getDeviceInfo() {
-    return from(Device.getInfo());
+    return forkJoin({
+      deviceInfo: Device.getInfo(),
+      deviceId: Device.getId(),
+      appInfo: this.getAppInfo(),
+    }).pipe(
+      map(({ deviceInfo, deviceId, appInfo }) =>
+        Object.assign(deviceInfo, deviceId, {
```
Julias0, line 21: can you map individual properties that we are using instead of using Object assign. Otherwise this piece of code is difficult to userstate.

Alternative: You can also choose to create a return type for `getDeviceInfo` method

ios/App/App/capacitor.config.json
```diff
@@ -1,14 +1,12 @@
...
-  "npmClient": "npm",
-  "webDir": "www",
-  "plugins": {
-    "SplashScreen": {
-      "launchShowDuration": 5000,
-      "androidScaleType": "CENTER_CROP"
-    }
-  },
-  "cordova": {}
+	"appId": "com.ionicframework.fyle595781",
```
Julias0, line 2: indentation should be 2 spaces - this looks different

Julias0 (approved):

ios/App/App/capacitor.config.json
yksolanki9: It is 2 spaces in vscode, looks like a bug in github

src/app/core/services/device.service.ts
yksolanki9: Done

PR #1066, opened 2022-05-03
Added unit test for accounts service api call method

Julias0:


src/app/core/services/accounts.service.ts
```diff
@@ -15,12 +15,8 @@ export class AccountsService {
     private currencyPipe: CurrencyPipe
   ) {}
 
-  getEMyAccounts(filters?) {
-    const data = {
-      params: filters,
-    };
-
-    return this.apiService.get('/eaccounts/', data).pipe(
+  getEMyAccounts() {
```
Julias0, line 18: filters were not used anywhere

PR #1081, opened 2022-05-09
Added code coverage for project service

Julias0:


src/app/core/services/projects.service.ts
```diff
@@ -91,20 +92,7 @@ export class ProjectsService {
     }
   }
 
-  filterById(projectId, projects) {
```
Julias0, line 94: not used

PR #1083, opened 2022-05-10
Cost center unit test

Julias0:


src/app/core/services/cost-centers.service.ts
```diff
@@ -18,28 +18,28 @@ export class CostCentersService {
...
-      .get('/cost_centers', data)
-      .pipe(map((res: PlatformCostCenter) => this.transformFrom(res.data)));
+      .get<PlatformCostCenter>('/cost_centers', data)
+      .pipe(map((res) => this.transformFrom(res.data)));
   }
 
-  transformFrom(platformCostCenter: PlatformCostCenterData[]): CostCenter[] {
-    let oldCostCenter = [];
+  private transformFrom(platformCostCenter: PlatformCostCenterData[]): CostCenter[] {
+    let oldCostCenter: CostCenter[] = [];
```
Dimple16, line 33: I guess we can skip this and directly use `const oldCostCenter` since we aren't resetting it

Dimple16 (approved):

PR #1090, opened 2022-05-10
Audit trail changes

saitejasFyle:


yksolanki9 (approved):

src/app/core/services/status.service.ts
```diff
@@ -77,6 +77,12 @@ export class StatusService {
           icon: 'edit',
         };
         break;
+      case lowerCaseComment.indexOf('policy violation will trigger the following action') > -1:
```
Julias0, line 80: This means older versions of mobile app while live will show wrong info. Will we be deprecating older versions just a few days after release? Please open a thread on this

saitejasFyle: https://fylein.slack.com/archives/C017C400V37/p1652788025928059?thread_ts=1652421497.848339&cid=C017C400V37

PR #1098, opened 2022-05-13
changed ratio of currency amount to 30:70

sandy-fyle:
changed ratio of currency amount to 30:70
image

image

image


Dimple16 (changes requested):
change looks good to me, can you add before and after screenshots?
also, test this across different resolutions

Dimple16 (approved):

PR #1100, opened 2022-05-15
changed the font and color of amount placeholder and text

sandy-fyle:
image

image
-----------------------------------------------------------------------------------------------------------------------------------------
image

image

### DIFF
--- a/src/app/shared/components/fy-number/fy-number.component.scss
+++ b/src/app/shared/components/fy-number/fy-number.component.scss
@@ -3,15 +3,17 @@
 .fy-number {
   &--input {
     border: 0;
-    font-size: 14px;
-    font-weight: 400;
+    font-size: 20px;
+    font-weight: 600;
     height: 24px;
     line-height: 18px;
     color: $blue-black;
     width: 100%;
     margin: 4px 0;
     padding: 0;
-
+    &::placeholder {
+      color: $blue-black !important;
+    }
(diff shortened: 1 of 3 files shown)


src/app/shared/components/fy-number/fy-number.component.scss
```diff
@@ -3,15 +3,17 @@
 .fy-number {
   &--input {
     border: 0;
-    font-size: 14px;
-    font-weight: 400;
+    font-size: 20px;
+    font-weight: 600;
```
Dimple16, line 7: 500?

Dimple16: check what Aditi had mentioned in the doc

Dimple16 (changes requested):

src/app/shared/components/fy-number/fy-number.component.scss
sandy-fyle: in figma design, 400 is mentioned but with font weight 400 it doesn't look the same as notion doc ss

Dimple16: 500 should do, as this is font-weight - `medium`

Dimple16 (changes requested):

Dimple16 (approved):

PR #1130, opened 2022-05-24
Fyle f1u9et v2

Aditya-shrivastav:
- Removed fy-currency component from add-per-diem and add-edit-expense component.
- Moved fy-currency component to shared folder for add-edit-expense.


src/app/shared/components/fy-currency/fy-currency.component.ts
```diff
@@ -11,7 +11,7 @@ import { CurrencyService } from '../../../core/services/currency.service';
 import { ModalPropertiesService } from 'src/app/core/services/modal-properties.service';
 
 @Component({
-  selector: 'app-fy-currency',
+  selector: 'app-new-fy-currency',
```
OmkarJ13, line 14: Why is `new` here?

Aditya-shrivastav: Since, we have app-fy-currency for advance, we cannot use the same selector name for shared component

OmkarJ13 (approved):

src/app/fyle/add-edit-expense/add-edit-expense.page.html
```diff
@@ -200,15 +200,15 @@
           <div>
             <div *ngIf="etxn$|async as etxn" class="add-edit-expense--currency">
               <ng-container *ngIf="homeCurrency$|async as homeCurrency">
-                <app-fy-currency
+                <app-new-fy-currency
```
rvab, line 203: Where is `app-new-fy-currency` component?

rvab (changes requested):
Can you update the description?. Please add why fy-currency component is removed from `add-per-diem` and `add-edit-expense` component.

rvab (approved):

src/app/fyle/add-edit-expense/add-edit-expense.page.html
Julias0: Why create a component called new fy component? Why not retain the same name or make changes inside the same component?

Julias0 (approved):

PR #1145, opened 2022-05-31
Fix modals `swipeToClose` deprecation warnings

yksolanki9:
Issue:
- We're getting `swipeToClose has been deprecated in favour of canDismiss` warnings after upgrading to ionic 6
- Ionic has deprecated swipeToClose property and it will be removed in v7. We use full page modals in our app and going forward full page modals cannot be closed by swiping.
- This comment from one of the maintainers summarises the point -  `The current thinking is that modals should only be swipeable when either using the card modal or the sheet modal. The fact that non-card/sheet modals could be swipeable was a mistake on our end. This change is being done to better align with native iOS and Android behavior`
- The docs for modals are not updated and all examples on the ionic website throw the same warning we’re getting

Solution
- I have converted all full page modals to sheet modals, there is no visual change, only the way we create a modal has changed and we still have the swipe to close functionality


Dimple16 (approved):

src/app/core/services/modal-properties.service.ts
```diff
@@ -11,9 +11,12 @@ export class ModalPropertiesService {
     const properties = {
       cssClass: 'fy-modal',
       showBackdrop: true,
-      swipeToClose: true,
+      canDismiss: true,
       backdropDismiss: true,
       animated: true,
+      initialBreakpoint: 1,
+      breakpoints: [0, 1],
+      handle: false,
```
rvab, line 19: Can we not move these options to `ModalPropertiesService`?  May be a create new method

src/app/fyle/add-edit-advance-request/add-edit-advance-request.page.ts
```diff
@@ -430,7 +429,6 @@ export class AddEditAdvanceRequestPage implements OnInit {
         canEdit: true,
       },
       mode: 'ios',
-      presentingElement: await this.modalController.getTop(),
```
rvab, line 433: Why remove it?

rvab, line 433: Why remove it?

src/global.scss
```diff
@@ -708,7 +708,7 @@ ref - https://stackoverflow.com/a/31162426
 }
 
 .fy-action-sheet {
-  --background: none;
+  --background: $white;
```
rvab, line 711: Can we not use `background-color`?

rvab (changes requested):

src/app/fyle/add-edit-advance-request/add-edit-advance-request.page.ts
yksolanki9: It is used when presenting modal on top of modal. Since we're presenting on top of page, it is redundant here

src/global.scss
yksolanki9: Action sheet is not a native html element, usual css properties do not directly work on it
https://ionicframework.com/docs/api/action-sheet

src/app/core/services/modal-properties.service.ts
rvab: The same custom properties are being used at 6 different places. Shouldn't we avoid this?. Can't we create a new method that returns the default + these new properties?

rvab (changes requested):

rvab (approved):

Julias0 (approved):

PR #1154, opened 2022-06-03
FYLE-29pu6en-2: Platform changes - Category service

Dimple16:
### New changes

1. Subcategory fix
- As the sub-category is null for parent-only categories, added the fix to work similarly on the public side 
`sub_category: category.sub_category || category.name`

2. Cluster domain fix
- SpenderPlatformApiService - cluster domain set root was missing in router-auth service
`this.spenderPlatformApiService.setRoot(domain);`


src/app/core/models/platform/platform-category.model.ts
```diff
@@ -0,0 +1,13 @@
+export interface PlatformCategory {
+  id: number;
+  org_id: string;
+  created_at: Date;
+  updated_at: Date;
+  name: string;
+  sub_category: string;
```
rvab, line 7: Shouldn't this be optional?

src/app/core/services/categories.service.ts
```diff
@@ -12,13 +15,66 @@ const categoriesCacheBuster$ = new Subject<void>();
...
+      reduce((acc, curr) => acc.concat(curr), [] as OrgCategory[])
+    );
+  }
+
+  getActiveCategoriesCount(): Observable<number> {
+    const data = {
+      params: {
+        is_enabled: 'eq.' + true,
+        offset: 0,
+        limit: 1,
```
rvab, line 40: only 1?

rvab (changes requested):

src/app/core/services/categories.service.ts
Dimple16: yes, this is just to get the count

src/app/core/models/platform/platform-category.model.ts
Dimple16: no, we'll have the key always

rvab (approved):

src/app/core/services/categories.service.ts
```diff
@@ -12,13 +15,66 @@ const categoriesCacheBuster$ = new Subject<void>();
...
-  getAll() {
-    return this.apiService.get('/org_categories').pipe(map(this.sortCategories), map(this.addDisplayName));
+  getAll(): Observable<OrgCategory[]> {
+    return this.getActiveCategoriesCount().pipe(
+      switchMap((count) => {
+        count = count > 50 ? count / 50 : 1;
+        return range(0, count);
+      }),
+      concatMap((page) => this.getCategories({ offset: 50 * page, limit: 50 })),
+      map((res) => res),
```
Julias0, line 30: this map dosnt do anything

src/app/core/services/categories.service.ts
```diff
@@ -12,13 +15,66 @@ const categoriesCacheBuster$ = new Subject<void>();
...
+    const data = {
+      params: {
+        is_enabled: 'eq.' + true,
+        offset: config.offset,
+        limit: config.limit,
+      },
+    };
+    return this.spenderPlatformApiService.get<PlatformApiResponse<PlatformCategory>>('/categories', data).pipe(
+      map((res) => this.transformFrom(res.data)),
+      map(this.sortCategories),
```
Julias0, line 58:
```
map(res => this.sortCategories(res)),
map(res => this.addDisplayName(res))

```

Passing a function as an argument messes up the reference of this associated with the function. Safer to use inside a lambda

Julias0 (approved):

PR #1181, opened 2022-06-16
Update other dependencies in mobile app

yksolanki9:


src/app/fyle/my-view-advance-request/my-view-advance-request.page.scss
```diff
@@ -155,12 +155,6 @@
     padding: 8px 0px;
   }
 
-  &--row-items {
```
Julias0, line 158: why are these changes here? How are these related to dependencies?

Julias0 (changes requested):
minor questions

src/app/fyle/my-view-advance-request/my-view-advance-request.page.scss
yksolanki9: Saw these warnings when building the app, so fixed them

yksolanki9: Created a separate pr for them

Julias0 (approved):

PR #1189, opened 2022-06-21
Add device info in mixpanel events

yksolanki9:
Sample values for my device

```
manufacturer:"OnePlus"
model:"AC2001"
operatingSystem:"android"
OVersion:"11"
platform:"android"
webViewVersion:"102.0.5005.125"
```

We can get the actual device name from the manufacturer + model details
Added the os, platform and webview version details just for additional debugging
I've also added app version to all mixpanel events, so we dont need to refer the db everytime

### DIFF
--- a/src/app/core/services/tracking.service.ts
+++ b/src/app/core/services/tracking.service.ts
@@ -1,6 +1,7 @@
 import { Injectable } from '@angular/core';
 import { AuthService } from './auth.service';
 import { DeviceService } from '../../core/services/device.service';
+import { environment } from 'src/environments/environment';
 
 @Injectable({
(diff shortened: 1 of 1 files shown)


src/app/core/services/tracking.service.ts
```diff
@@ -64,6 +64,15 @@ export class TrackingService {
         ...properties,
         Asset: 'Mobile',
         DeviceType: deviceInfo.platform,
```
yksolanki9, line 66: This property is repeated in the deviceInfo object, but I'm not removing it from here as it may affect the existing reports

Julias0 (approved):

PR #1190, opened 2022-06-21
Virtual scroll in select modal

Dimple16:


Julias0:
The scroll should include the recently selected items too - otherwise the interaction looks a bit weired

Julias0 (changes requested):
The scroll should include the recently selected items too - otherwise the interaction looks a bit weired

Dimple16:
@Julias0 the recently used is a separate list, in this case, so we would need to add 2 virtual scrolls, I tried this too, since max only 4 items can be present, the scroll doesn't have an impact as such here

Julias0:
> @Julias0 the recently used is a separate list, in this case, so we would need to add 2 virtual scrolls, I tried this too, since max only 4 items can be present, the scroll doesn't have an impact as such here

Cant we concatenate both the lists so that it feels like they are part of the same?

Julias0 (changes requested):
Asked questions on slack

Julias0 (changes requested):
This implementation needs to be changed

PR #1194, opened 2022-06-22
Resolved undefined error using optional chaining

saitejasFyle:


src/app/core/services/deep-link.service.ts
```diff
@@ -10,7 +10,7 @@ export class DeepLinkService {
   getJsonFromUrl(url) {
     const query = url?.split('?')[1];
     const result = {};
-    query.split('&').forEach((part) => {
+    query?.split('&').forEach((part) => {
```
Julias0, line 13: In what case would query be not present? Is that edge case ka behaviour handled?

Julias0 (changes requested):
added comments

src/app/core/services/deep-link.service.ts
saitejasFyle: Yup, the edge cases are handled already. When the URL parts are not present, we'll be redirecting the users to switch organisation page

Julias0 (approved):

PR #1203, opened 2022-06-27
Removed number filter + slice hack

OmkarJ13:


abhishek1234321 (approved):

src/app/fyle/add-edit-expense/match-transaction/match-transaction.component.html
```diff
@@ -36,9 +36,7 @@
           {{ cccExpense.vendor || cccExpense.description | ellipsis: 20 }}
         </div>
         <div class="match-transaction--amount">
-          <span class="match-transaction--currency">{{
-            0 | currency: cccExpense.currency:'symbol-narrow' | slice: 0:1
-          }}</span
+          <span class="match-transaction--currency">{{ getCurrencySymbol(cccExpense.currency) }}</span
```
Julias0, line 39: we shouldnt use method calls inside the template

Julias0 (changes requested):
Comment added

src/app/fyle/add-edit-expense/match-transaction/match-transaction.component.html
```diff
@@ -36,7 +36,7 @@
           {{ cccExpense.vendor || cccExpense.description | ellipsis: 20 }}
         </div>
         <div class="match-transaction--amount">
-          <span class="match-transaction--currency">{{ getCurrencySymbol(cccExpense.currency) }}</span
+          <span class="match-transaction--currency">{{ cccExpense.currency | currencySymbol }}</span
```
OmkarJ13, line 39: - Added this new pipe which will return a symbol for a currency.
- This was done to avoid setting currency symbols in component.ts to use inside of template files or calling methods which return symbols directly from template files.

src/app/shared/components/fy-currency/fy-currency.component.html
```diff
@@ -20,7 +20,7 @@
   <div class="fy-currency--exchange-rate-container" [ngClass]="{ expanded: expanded }">
     <mat-icon class="fy-currency--exchange-rate-icon" svgIcon="fy-info"></mat-icon>
     <div class="fy-currency--exchange-rate-message">
-      <strong>{{ value.currency }} {{ value.amount | number: '1.0-2' }}</strong>
+      <strong>{{ value.amount | currency: value.currency:'code' }}</strong>
```
Julias0, line 23: What is the `code` value getting passed here?
In the pipes transform function the only two valid options given are `wide` and `narrow`

Julias0 (changes requested):
Added questions

src/app/shared/components/fy-currency/fy-currency.component.html
OmkarJ13: - This is done because we want to have the currency code as the symbol itself
- This is valid because in currency pipe we accept these options - `symbol` (same as wide), `symbol-narrow`, `code` 
- The pipe `currencySymbol` that I implemented in this PR is different, it only returns the symbol for the currency. There only two options are there which you have mentioned.
- I had implemented a different pipe for transforming currency code to a symbol because of the issue where we had to get the currency symbol using a method and use it in the template.

Julias0 (approved):

PR #1251, opened 2022-07-20
Fixed spacing problems with OMR currency

OmkarJ13:
After the fix
image
image


src/app/fyle/team-reports/team-report-card/team-report-card.component.html
```diff
@@ -21,7 +21,7 @@
           </div>
         </ion-col>
         <ion-col size="4">
-          <div class="text-right">
+          <div class="text-right reports-card--amount-box">
```
OmkarJ13, line 24: Added this to add `margin-right` to whole amount instead of just the number. It will look better than before

abhishek1234321 (approved):

PR #1264, opened 2022-07-22
Transaction outbox tests

Julias0:


src/app/core/services/transactions-outbox.service.spec.ts
```diff
@@ -1,16 +1,611 @@
...
+import { ReportService } from './report.service';
+import { StatusService } from './status.service';
+import { StorageService } from './storage.service';
+import { TrackingService } from './tracking.service';
+import { TransactionService } from './transaction.service';
 
 import { TransactionsOutboxService } from './transactions-outbox.service';
 
-xdescribe('TransactionsOutboxService', () => {
+const dataExtractionQueueBulkFyleSample = [
```
Julias0, line 16: This is a sample I picked up from staging

src/app/core/services/transactions-outbox.service.ts
```diff
@@ -62,13 +61,8 @@ export class TransactionsOutboxService {
...
   }
 
-  async removeDataExtractionEntry(expense, dataUrls) {
-    const entry = {
-      transaction: expense,
-      dataUrls,
-    };
-
-    const idx = this.dataExtractionQueue.indexOf(entry);
+  async removeDataExtractionEntry(expenseId: string) {
```
Julias0, line 64: The older implementation did not work, idx will always be -1 in the old implementation. 
THe only reason this works in production is that we keep a clone of the original array when processing the items for data extraction when app is running. If app is closed in middle of data extraction, some items will remain without data extraction

PR #1276, opened 2022-07-27
Upgrade Android API level to 31

Dimple16:


android/app/src/main/AndroidManifest.xml
```diff
@@ -18,6 +18,7 @@
             android:theme="@style/AppTheme.NoActionBarLaunch"
             android:launchMode="singleTask"
             android:screenOrientation="portrait"
+            android:exported="true"
```
yksolanki9, line 21: Why do we need this?

yksolanki9 (approved):

android/app/src/main/AndroidManifest.xml
Julias0: +1

Julias0 (changes requested):
1 question

android/app/src/main/AndroidManifest.xml
Dimple16: @Julias0 
We need to explicitly set this for app targeting android 12 and above. adding references:
https://developer.android.com/guide/topics/manifest/activity-element#exported
https://developer.android.com/about/versions/12/behavior-changes-12
https://stackoverflow.com/questions/68554294/androidexported-needs-to-be-explicitly-specified-for-activity-apps-targeting

Julias0 (approved):

PR #1285, opened 2022-08-05
#payment_mode_configurations - Account type enum

yksolanki9:
Created enum for account types and added type for expense types


Julias0 (approved):

src/app/core/services/accounts.service.ts
```diff
@@ -114,7 +115,7 @@ export class AccountsService {
     accounts: ExtendedAccount[],
     orgSettings: any,
     etxn?: any,
-    expenseType = 'EXPENSE'
+    expenseType: 'EXPENSE' | 'MILEAGE' | 'PER_DIEM' = 'EXPENSE'
```
yksolanki9, line 118: Not creating an enum for this as it is used only in one place

gokulk16 (approved):

PR #1293, opened 2022-08-19
Added fix for virtual scroll

Julias0:


src/app/shared/components/virtual-select/virtual-select-modal/virtual-select-modal.component.html
```diff
@@ -0,0 +1,104 @@
...
+    </mat-form-field>
+  </ion-toolbar>
+</ion-header>
+
+<ion-content>
+  <cdk-virtual-scroll-viewport
+    class="selection-modal--virtual-scroll"
+    itemSize="56"
+    minBufferPx="900"
+    maxBufferPx="1350"
```
rvab, line 38: any logical reason to keep the number as 900 and 1350?

src/app/shared/components/virtual-select/virtual-select-modal/virtual-select-modal.component.ts
```diff
@@ -0,0 +1,207 @@
...
+              option.selected = isEqual(option.value, this.currentSelection);
+              return option;
+            })
+        )
+      );
+    }
+  }
+
+  ngAfterViewInit() {
+    if (this.searchBarRef && this.searchBarRef.nativeElement) {
```
rvab, line 93: 1. Can you split this method?
2. It's better to run these heavy lifting task outside of angular zone.

src/app/shared/components/virtual-select/virtual-select-modal/virtual-select-modal.component.ts
```diff
@@ -0,0 +1,207 @@
...
+        )
+      );
+    } else {
+      const initial = [];
+
+      if (this.nullOption && this.currentSelection) {
+        initial.push({ label: 'None', value: null });
+      }
+
+      this.recentlyUsedItems$ = of([]);
```
rvab, line 147: Why this is an empty array?

rvab (approved):

src/app/shared/components/virtual-select/virtual-select-modal/virtual-select-modal.component.ts
Julias0: because `this.recentlyUsedItems$` is unset in this case. I have to set it to an observable for it to work with combineLatest which I have used later

src/app/shared/components/virtual-select/virtual-select-modal/virtual-select-modal.component.ts
Julias0: Ah, most of the processing is happening inside the observable pipelines, so not much code execution will occur here.

Also I have just duplicated this from fy-select, I will leave refactoring of old code for later

Dimple16 (approved):

PR #1298, opened 2022-08-19
Remove offline forms support

Dimple16:


src/app/app.component.ts
```diff
@@ -66,7 +67,8 @@ export class AppComponent implements OnInit {
     private loginInfoService: LoginInfoService,
     private offlineService: OfflineService,
     private navController: NavController,
-    private popoverController: PopoverController
+    private popoverController: PopoverController,
+    private removeOfflineFormsService: RemoveOfflineFormsService
```
Julias0, line 71: remove from here since we are not using anymore

src/app/core/services/remove-offline-forms.service.ts
```diff
@@ -0,0 +1,57 @@
...
+    });
+  }
+
+  getRemoveOfflineFormsLDKey() {
+    const eou$ = from(this.authService.getEou());
+    const currentOrg$ = this.offlineService.getCurrentOrg().pipe(shareReplay(1));
+    const deviceInfo$ = this.deviceService.getDeviceInfo().pipe(shareReplay(1));
+    const deleteLDKey$ = from(this.storageService.delete('isOfflineFormsRemoved'));
+    return forkJoin([eou$, deviceInfo$, currentOrg$, deleteLDKey$]).pipe(
+      timeout(2000),
```
Julias0, line 40: add the timeout after the switchMap

src/app/core/services/remove-offline-forms.service.ts
```diff
@@ -0,0 +1,57 @@
...
+            const LDClient = this.launchDarklyService.getLDClient();
+            LDClient.waitForInitialization().then(() => {
+              const allLDFlags = LDClient.allFlags();
+              // eslint-disable-next-line @typescript-eslint/dot-notation
+              subscriber.next(allLDFlags['remove_offline_forms']);
+              subscriber.complete();
+            });
+          })
+      ),
+      catchError((err) => of(null))
```
Julias0, line 54: instead of null, let's explicitly return false.

Julias0 (approved):

PR #1299, opened 2022-08-23
Add ld flag to disable keyboard plugin

yksolanki9:
- BR link - https://app.clickup.com/t/2fkq1y2
- Tested these changes on iphone 13 🟢  for both cases locally
- This will allow all users in regions with period as decimal separator(includes USD) to use third party keyboards, but users in other regions will still have to use ios default keyboard
- Will add a flag in LD after this pr is approved

### DIFF
--- a/src/app/core/services/launch-darkly.service.ts
+++ b/src/app/core/services/launch-darkly.service.ts
@@ -47,6 +47,10 @@ export class LaunchDarklyService {
     this.ldClient.on('change', this.updateCache, this);
   }
 
+  checkIfKeyboardPluginIsDisabled() {
+    return this.getVariation('disable_keyboard_plugin', false);
+  }
+
   private updateCache() {
     if (this.ldClient) {
       const latestFlags = this.ldClient.allFlags();
--- a/src/app/core/services/launch-darkly.service.ts
+++ b/src/app/core/services/launch-darkly.service.ts
@@ -47,8 +47,8 @@ export class LaunchDarklyService {
     this.ldClient.on('change', this.updateCache, this);
   }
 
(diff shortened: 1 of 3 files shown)


src/app/core/services/launch-darkly.service.ts
```diff
@@ -47,6 +47,10 @@ export class LaunchDarklyService {
     this.ldClient.on('change', this.updateCache, this);
   }
 
+  checkIfKeyboardPluginIsDisabled() {
+    return this.getVariation('disable_keyboard_plugin', false);
```
Julias0, line 51: I think it would be better if we have `keyboard_plugin_enabled`.
Disabled flags are always hard to read

yksolanki9: Fixed

src/app/shared/components/fy-number/fy-number.component.html
```diff
@@ -1,5 +1,5 @@
 <input
-  *ngIf="isIos"
+  *ngIf="isIos && isKeyboardPluginEnabled; else inputWithoutPlugin"
```
Julias0, line 2: Are we sure using the ngIf inside input will not cause a nested input statement? I would be happier if we moved the ngIf into a `ng-container`

Julias0 (changes requested):
Minor comments

src/app/shared/components/fy-number/fy-number.component.html
yksolanki9: Done

Julias0 (approved):

PR #1302, opened 2022-08-23
Fix ios camera preview preview not working sometimes

yksolanki9:
- BR link - https://app.clickup.com/t/2np4ufy
- The issue is because, `ngAfterViewInit()` is not getting called sometimes when we use instafyle, save the expense and open the camera again. 
- I've replaced angular llifecycle hooks with the ones provided by ionic - `ionViewWillEnter()` and `ionViewWillLeave()`. This is what ionic recommends as well - [Ref](https://ionicframework.com/docs/angular/lifecycle#:~:text=When%20an%20app,a%20page%20%22popped%22.)

### DIFF
--- a/src/app/shared/components/capture-receipt/capture-receipt.component.ts
+++ b/src/app/shared/components/capture-receipt/capture-receipt.component.ts
@@ -1,4 +1,4 @@
-import { Component, EventEmitter, OnDestroy, OnInit, Input, AfterViewInit } from '@angular/core';
+import { Component, EventEmitter, OnInit, Input } from '@angular/core';
 import { CameraPreview, CameraPreviewOptions, CameraPreviewPictureOptions } from '@capacitor-community/camera-preview';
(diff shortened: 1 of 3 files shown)


src/app/shared/components/capture-receipt/capture-receipt.component.ts
```diff
@@ -28,7 +28,7 @@ type Image = Partial<{
   templateUrl: './capture-receipt.component.html',
   styleUrls: ['./capture-receipt.component.scss'],
 })
-export class CaptureReceiptComponent implements OnInit, OnDestroy, AfterViewInit {
+export class CaptureReceiptComponent implements OnInit {
```
Julias0, line 31: what about the case where we open this inside a modal? Will `ionViewWillEnter` also be triggered then?

Julias0 (approved):

PR #1318, opened 2022-09-05
Sentry fixes in dashboard and tasks pages

Dimple16:
- TypeError: Cannot read properties of null (reading 'corporate_credit_card_settings')
- TypeError: Cannot read properties of null (reading 'length')
- TypeError: Cannot read properties of null (reading 'approval_settings')

https://sentry.io/organizations/fyle-technologies-private-limi/issues/3409776814/?project=5622998&query=is%3Aunresolved&statsPeriod=14d

### DIFF
--- a/src/app/core/services/tasks.service.ts
+++ b/src/app/core/services/tasks.service.ts
@@ -445,19 +445,14 @@ export class TasksService {
   }
 
   private getTeamReportsStats() {
-    return forkJoin({
-      eou: from(this.authService.getEou()),
-      sequentalApproversEnabled: this.offlineService
-        .getOrgSettings()
-        .pipe(map((orgSettings) => orgSettings?.approval_settings?.enable_sequential_approvers)),
-    }).pipe(
-      switchMap(({ eou, sequentalApproversEnabled }) =>
+    return from(this.authService.getEou()).pipe(
+      switchMap((eou) =>
         this.reportService.getReportStatsData(
           {
             approved_by: 'cs.{' + eou.ou.id + '}',
             rp_approval_state: ['in.(APPROVAL_PENDING)'],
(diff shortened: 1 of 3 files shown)


src/app/core/services/tasks.service.ts
```diff
@@ -449,7 +449,7 @@ export class TasksService {
       eou: from(this.authService.getEou()),
       sequentalApproversEnabled: this.offlineService
```
Julias0, line 450: I dont think `sequentalApproversEnabled` is used at all. You can remove this off if that is the case

src/app/fyle/dashboard/tasks/tasks.component.ts
```diff
@@ -453,7 +453,7 @@ export class TasksComponent implements OnInit {
               eou: from(this.authService.getEou()),
               sequentalApproversEnabled: this.offlineService
```
Julias0, line 454: same comment as above

src/app/fyle/dashboard/tasks/tasks.component.ts
```diff
@@ -448,18 +448,10 @@ export class TasksComponent implements OnInit {
...
-          switchMap(() =>
-            forkJoin({
-              eou: from(this.authService.getEou()),
-              sequentalApproversEnabled: this.offlineService
-                .getOrgSettings()
-                .pipe(map((orgSettings) => orgSettings.approval_settings.enable_sequential_approvers)),
-            })
-          ),
-          map(({ eou, sequentalApproversEnabled }) => ({
+          map(() => ({
```
yksolanki9, line 451: Why do we need map here? We can define and use the object in the next operator directly

yksolanki9 (changes requested):

yksolanki9 (approved):

PR #1323, opened 2022-09-07
#automate_report_submission: Add LD flag in expense form and tasks page

yksolanki9:
Changes in this PR
- Added LD support
- Content change in add to report list view - Removed state pill and updated text

New design
localhost_8100_enterprise_add_edit_expense;navigate_back=true(iPhone SE) (2)

### DIFF
--- a/src/app/core/services/launch-darkly.service.ts
+++ b/src/app/core/services/launch-darkly.service.ts
@@ -62,7 +62,7 @@ export class LaunchDarklyService {
     return this.getVariation('keyboard_plugin_enabled', true);
   }
 
-  checkIfReportAutoSubmissionIsEnabled() {
+  checkIfAutomateReportSubmissionIsEnabled() {
     return this.getVariation('automate_report_submission_enabled', false);
   }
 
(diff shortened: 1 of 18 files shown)


src/app/core/services/launch-darkly.service.ts
```diff
@@ -62,6 +62,10 @@ export class LaunchDarklyService {
     return this.getVariation('keyboard_plugin_enabled', true);
   }
 
+  checkIfReportAutoSubmissionIsEnabled() {
```
Dimple16, line 65: lets have the method name same as the key -> `checkIfAutomateReportSubmissionIsEnabled`

Dimple16: also, why do we have the key with `_enabled`? isn't this incorrect? `automate_report_submission_enabled` this should have been `automate_report_submission`

Dimple16 (changes requested):

src/app/core/services/launch-darkly.service.ts
yksolanki9: Yes, `enabled` does sound a bit redundant here, but this is being used across all clients and will be removed when we release this feature for all orgs, so shouldnt be an issue

yksolanki9: Renamed the method
\

Dimple16 (approved):

PR #1329, opened 2022-09-08
FYLE-2tea9p4: Increase chunk size to 200 from 50 for APIs

Dimple16:


src/app/core/services/cost-centers.service.ts
```diff
@@ -21,10 +21,10 @@ export class CostCentersService {
   getAllActive(): Observable<CostCenter[]> {
     return this.getActiveCostCentersCount().pipe(
       switchMap((count) => {
-        count = count > 50 ? count / 50 : 1;
+        count = count > 200 ? count / 200 : 1;
```
Julias0, line 24: cant we use the injected `PAGINATION_SIZE` instead of the hardcoded value everywhere?
Next time if we want to change this, we wont have to change this at every place

Julias0 (changes requested):
Added comments

src/app/core/services/cost-centers.service.ts
OmkarJ13: heavy plus 👍

yksolanki9 (approved):
Minor comments

Julias0 (approved):

PR #1338, opened 2022-09-12
Tracking switch org launch time

Dimple16:
Screenshot 2022-09-12 at 4 14 31 PM

### DIFF
--- a/src/app/auth/sign-in/sign-in.page.ts
+++ b/src/app/auth/sign-in/sign-in.page.ts
@@ -66,7 +66,6 @@ export class SignInPage implements OnInit {
           take(1),
           switchMap(() => this.authService.refreshEou()),
           tap(async () => {
-            await this.trackLoginInfo();
             const markOptions: PerformanceMarkOptions = {
               detail: 'SAML Login',
             };
@@ -74,6 +73,7 @@ export class SignInPage implements OnInit {
             this.trackingService.onSignin(this.fg.value.email, {
               label: 'Email',
             });
+            await this.trackLoginInfo();
           })
         )
         .subscribe(() => {
@@ -183,7 +183,6 @@ export class SignInPage implements OnInit {
           }),
           switchMap(() => this.authService.refreshEou()),
           tap(async () => {
-            await this.trackLoginInfo();
             const markOptions: PerformanceMarkOptions = {
               detail: 'Password Login',
             };
@@ -191,6 +190,7 @@ export class SignInPage implements OnInit {
             this.trackingService.onSignin(this.fg.value.email, {
               label: 'Email',
             });
+            await this.trackLoginInfo();
           }),
           finalize(() => (this.passwordLoading = false))
         )
@@ -227,7 +227,6 @@ export class SignInPage implements OnInit {
             }),
             switchMap((res) => this.authService.refreshEou()),
             tap(async () => {
-              await this.trackLoginInfo();
               const markOptions: PerformanceMarkOptions = {
                 detail: 'Google Login',
               };
@@ -235,6 +234,7 @@ export class SignInPage implements OnInit {
               this.trackingService.onSignin(this.fg.value.email, {
                 label: 'Email',
               });
+              await this.trackLoginInfo();
             })
           )
         ),
--- a/src/app/auth/sign-in/sign-in.page.ts
+++ b/src/app/auth/sign-in/sign-in.page.ts
@@ -67,6 +67,10 @@ export class SignInPage implements OnInit {
           switchMap(() => this.authService.refreshEou()),
           tap(async () => {
             await this.trackLoginInfo();
+            const markOptions: PerformanceMarkOptions = {
+              detail: 'SAML Login',
+            };
+            performance.mark('login start time', markOptions);
             this.trackingService.onSignin(this.fg.value.email, {
               label: 'Email',
             });
@@ -180,6 +184,10 @@ export class SignInPage implements OnInit {
           switchMap(() => this.authService.refreshEou()),
           tap(async () => {
             await this.trackLoginInfo();
+            const markOptions: PerformanceMarkOptions = {
+              detail: 'Password Login',
+            };
+            performance.mark('login start time', markOptions);
             this.trackingService.onSignin(this.fg.value.email, {
               label: 'Email',
             });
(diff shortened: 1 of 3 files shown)


src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -97,6 +97,22 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
         .pipe(switchMap(() => from(that.proceed())))
         .subscribe(noop);
     } else {
+      if (performance.getEntriesByName('switch org launch time').length < 1) {
+        // Time taken to land on switch org page after sign-n
+        performance.mark('switch org launch time');
+
+        // Measure total time taken from launching the app to landing on switch org page
+        performance.measure('switch org launch time', 'app launch start time', 'switch org launch time');
```
Julias0, line 105: Shouldnt we start this when user clicks on login?
`app launch start time` happens inside app.component.ts which would execute when user opens the app. For folks who are not logged in, it will also take into account the users interaction time while they are entering email and password

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -97,6 +97,26 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
         .pipe(switchMap(() => from(that.proceed())))
         .subscribe(noop);
     } else {
+      if (performance.getEntriesByName('switch org launch time').length < 1) {
```
Julias0, line 100: put this inside a try..catch just to be safe. And Extract this out into a separate funciton

Julias0 (changes requested):
Added minor comments - fix and merge

src/app/auth/sign-in/sign-in.page.ts
```diff
@@ -180,6 +184,10 @@ export class SignInPage implements OnInit {
           switchMap(() => this.authService.refreshEou()),
           tap(async () => {
             await this.trackLoginInfo();
+            const markOptions: PerformanceMarkOptions = {
+              detail: 'Password Login',
+            };
+            performance.mark('login start time', markOptions);
```
yksolanki9, line 190: Same comment as above

src/app/auth/sign-in/sign-in.page.ts
```diff
@@ -220,6 +228,10 @@ export class SignInPage implements OnInit {
             switchMap((res) => this.authService.refreshEou()),
             tap(async () => {
               await this.trackLoginInfo();
+              const markOptions: PerformanceMarkOptions = {
+                detail: 'Google Login',
+              };
+              performance.mark('login start time', markOptions);
```
yksolanki9, line 234: Same as above

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -282,4 +283,29 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
     this.contentRef.nativeElement.classList.remove('switch-org__content-container__content-block--hide');
     this.searchRef.nativeElement.classList.remove('switch-org__content-container__search-block--show');
   }
+
+  private trackSwitchOrgLaunchTime() {
+    try {
+      if (performance.getEntriesByName('switch org launch time').length < 1) {
```
yksolanki9, line 289: Use not operator or equate with 0?

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -282,4 +283,29 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
...
+        // Time taken to land on switch org page after sign-in
+        performance.mark('switch org launch time');
+
+        // Measure total time taken from logging into the app to landing on switch org page
+        performance.measure('switch org launch time', 'login start time', 'switch org launch time');
```
yksolanki9, line 294: You can skip the 3rd arg here, it'll anyways consider the current time.

yksolanki9: Also, should we should define mark names as enums. Strings are prone to errors

yksolanki9 (changes requested):

src/app/auth/switch-org/switch-org.page.ts
Dimple16: it won't make a difference, but, I'll equate to zero if that looks better :P

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -282,4 +283,29 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
...
+
+        const measureLaunchTime = performance.getEntriesByName('switch org launch time');
+
+        // eslint-disable-next-line @typescript-eslint/dot-notation
+        const loginMethod = performance.getEntriesByName('login start time')[0]['detail'];
```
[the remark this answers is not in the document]
Dimple16: since it uses a `PerformanceMarkOptions` interface, you cannot access it directly, we are forced to use it this way

src/app/auth/switch-org/switch-org.page.ts
Dimple16: enums are added in the next PR

src/app/auth/sign-in/sign-in.page.ts
```diff
@@ -67,6 +67,10 @@ export class SignInPage implements OnInit {
           switchMap(() => this.authService.refreshEou()),
           tap(async () => {
             await this.trackLoginInfo();
+            const markOptions: PerformanceMarkOptions = {
+              detail: 'SAML Login',
+            };
+            performance.mark('login start time', markOptions);
```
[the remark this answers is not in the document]
yksolanki9: Any reason we add this to the beginning of the `else` block?

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -97,6 +97,7 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
         .pipe(switchMap(() => from(that.proceed())))
         .subscribe(noop);
     } else {
+      that.trackSwitchOrgLaunchTime();
```
yksolanki9, line 100: Should we add this on line 123 when we're setting `isLoading` to false?

yksolanki9 (changes requested):

yksolanki9 (approved):

Julias0 (approved):

PR #1354, opened 2022-09-15
Removal of org settings, org user settings & accounts API call dependency from offline service

Dimple16:


src/app/core/services/accounts.service.ts
```diff
@@ -103,4 +104,42 @@ export class AccountsService {
       })
     );
   }
+
+  getAccount(
```
Julias0, line 108: This should be getDefaultAccount - also drop a message to Yash about this so that he is aware that he has to make changes here

src/app/core/services/transaction.service.ts
```diff
@@ -282,6 +289,11 @@ export class TransactionService {
           transaction.to_dt = this.timezoneService.convertToUtc(transaction.to_dt, offset);
         }
 
+        if (!transaction.source_account_id) {
+          transaction.source_account_id = txnAccount.source_account_id;
+          transaction.skip_reimbursement = txnAccount.skip_reimbursement;
```
Julias0, line 294: This looks a bit hacky - why is this happening here? This should happen in transactionOutbox

src/app/core/services/transaction.service.ts
```diff
@@ -962,4 +974,25 @@ export class TransactionService {
...
     return currentParamsCopy;
   }
+
+  getTxnAccount() {
+    return forkJoin({
+      orgUserSettings: this.offlineService.getOrgUserSettings(),
+      accounts: this.offlineService.getAccounts(),
+      orgSettings: this.offlineService.getOrgSettings(),
+    }).pipe(
+      switchMap(({ orgUserSettings, accounts, orgSettings }) =>
```
Julias0, line 984: witchMap(({ orgUserSettings, accounts, orgSettings }) =>this.accountsService.getAccount(orgSettings, accounts, orgUserSettings)),
 filter((account) => !!account), // why is this required? If account is null it wont emit anything and the parent forkJoin calling it.wont execute
 map((account) => {
            const accountDetails = {
              source_account_id: account.acc.id,
              skip_reimbursement: !account.acc.isReimbursable || false,
            };
            return accountDetails;
          })

Julias0 (changes requested):
Added comments

src/app/core/services/transaction.service.ts
Dimple16: @Julias0 Adding it to the transaction outbox creates a race condition where the source account ID is never set,
In this case, we add the account ID exactly while upsert if the account ID isn't present, 
This is the expected behavior for expense creations, none of the online created expenses will have the `source_account_id` as null, also, even if it is null, we want the account ID to be set if not present, it shouldn't be a problem imo,

Julias0: ok

Julias0 (approved):

src/app/shared/components/capture-receipt/capture-receipt.component.ts
```diff
@@ -118,85 +118,35 @@ export class CaptureReceiptComponent implements OnInit, OnDestroy, AfterViewInit
...
-            } else {
-              return this.transactionsOutboxService.addEntryAndSync(transaction, attachmentUrls, null, null);
-            }
-          })
-        )
-      )
-    );
-  }
+    return this.networkService.isOnline().pipe(
+      map((isConnected) => {
```
yksolanki9, line 122: Use `switchMap` in case the network is unstable

yksolanki9 (approved):
Minor comment

Julias0 (approved):

PR #1355, opened 2022-09-19
Sentry fix: Check for app version only if user is logged in

yksolanki9:
- Fixes https://sentry.io/organizations/fyle-technologies-private-limi/issues/2234550581/?project=5622998 and https://sentry.io/organizations/fyle-technologies-private-limi/issues/2244675010/?project=5622998
- Moved `checkAppSupportedVersion()` call to `platform.ready()` as device plugin may not always be initialized inside `ngOnInit()`
- Some refactoring


src/app/app.component.ts
```diff
@@ -136,43 +137,22 @@ export class AppComponent implements OnInit {
...
+          this.appVersionService.checkAppSupportedVersion(deviceInfo);
+        });
 
       // Global cache config
       GlobalCacheConfig.maxAge = 10 * 60 * 1000;
       GlobalCacheConfig.maxCacheCount = 100;
     });
   }
 
-  checkAppSupportedVersion() {
```
yksolanki9, line 147: Moved this method inside `app-version` service and added it inside switch org as well.

src/app/core/services/offline.service.ts
```diff
@@ -511,7 +509,10 @@ export class OfflineService {
     const delegatedAccounts$ = this.getDelegatedAccounts();
     const taxGroups$ = this.getEnabledTaxGroups();
 
-    this.loadAppVersion();
+    this.deviceService.getDeviceInfo().subscribe((deviceInfo) => {
+      this.loadAppVersion(deviceInfo);
+      this.appVersionService.checkAppSupportedVersion(deviceInfo);
```
yksolanki9, line 514: Added check for supported version in switch org page as well. This fixes a corner case where user is able to login to an deprecated app and create expense

src/app/app.component.ts
```diff
@@ -189,7 +169,6 @@ export class AppComponent implements OnInit {
         .forEach((key) => lstorage.removeItem(key));
     }
 
-    this.checkAppSupportedVersion();
```
yksolanki9, line 192: Moved this method inside `platform.ready()` as capacitor plugins may not be always initialized inside `ngOnInit`

src/app/app.component.ts
```diff
@@ -136,43 +137,22 @@ export class AppComponent implements OnInit {
       });
       setTimeout(async () => await SplashScreen.hide(), 1000);
 
-      this.offlineService.loadAppVersion();
+      from(this.routerAuthService.isLoggedIn())
+        .pipe(
+          filter((isLoggedIn) => !!isLoggedIn),
```
yksolanki9, line 142: `/mobileapp/check` and `version/app/<os>` calls should only be made if the user is logged in. Else these checks will be done  after login in switch org page

rvab: Why these calls should be made after the user is logged in? What happens if done before and Why?

rvab (changes requested):

src/app/app.component.ts
yksolanki9: I have already added the sentry issue link in the description. We'd get `403: Forbidden` as the user does not have a valid access token

src/app/core/services/app-version.service.ts
```diff
@@ -43,35 +50,52 @@ export class AppVersionService {
...
+          lastLoggedInVersion,
+          user_email: eou?.us?.email,
+          appVersion: deviceInfo.appVersion,
+        });
+        this.router.navigate(['/', 'auth', 'app_version', { message: appSupportDetails.message }]);
+      }
+    });
+  }
+
   isSupported(data) {
```
Julias0, line 99: isSupported(deviceInfo) {

src/app/core/services/app-version.service.ts
```diff
@@ -43,35 +50,52 @@ export class AppVersionService {
...
+      lastLoggedInVersion: this.loginInfoService.getLastLoggedInVersion(),
+      eou: from(this.authService.getEou()),
+    }).subscribe(({ appSupportDetails, lastLoggedInVersion, eou }) => {
+      if (!appSupportDetails.supported && environment.production) {
+        this.trackingService.eventTrack('Auto Logged out', {
+          lastLoggedInVersion,
+          user_email: eou?.us?.email,
+          appVersion: deviceInfo.appVersion,
+        });
+        this.router.navigate(['/', 'auth', 'app_version', { message: appSupportDetails.message }]);
```
Julias0, line 94: we shouldnt inject the router

Julias0: Move subscribe block to page level component

src/app/core/services/offline.service.ts
```diff
@@ -511,7 +509,10 @@ export class OfflineService {
     const delegatedAccounts$ = this.getDelegatedAccounts();
     const taxGroups$ = this.getEnabledTaxGroups();
 
-    this.loadAppVersion();
+    this.deviceService.getDeviceInfo().subscribe((deviceInfo) => {
```
Julias0, line 512: move the behaviour change to inside switch org page

Julias0 (approved):

src/app/app.component.ts
rvab: Why does the user need to be authorized to make this API? This check should be independent no? If the app version is deprecated, it's deprecated. Why depend on login status?

rvab (approved):
One clarification

src/app/app.component.ts
yksolanki9: I dont think we can allow any of our internal apis to be accessible without a token. Anyone can use a DDoS attack against us?

src/app/core/services/app-version.service.ts
yksolanki9: Fixed

src/app/core/services/app-version.service.ts
yksolanki9: Done

src/app/core/services/offline.service.ts
yksolanki9: Done

PR #1359, opened 2022-09-21
Fyle 2k5y2va: Added new guard

Aditya-shrivastav:


src/app/core/guards/org-user-auth.guard.ts
```diff
@@ -0,0 +1,24 @@
+import { Injectable } from '@angular/core';
+import { ActivatedRouteSnapshot, CanActivate, Router, RouterStateSnapshot, UrlTree } from '@angular/router';
+import { Observable } from 'rxjs';
+import { AuthService } from '../services/auth.service';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class OrgUserAuthGuard implements CanActivate {
```
Julias0, line 9: Lets call this VerifiedOrgAuthGuard

Julias0 (approved):

PR #1367, opened 2022-09-22
FYLE-2v1qxxu-1: Remove offline service methods - Currency and Home currency

Dimple16:


src/app/core/services/refiner.service.ts
```diff
@@ -242,7 +242,7 @@ export class RefinerService {
     return forkJoin({
       isConnected: this.isConnected$.pipe(take(1)),
       eou: this.authService.getEou(),
-      homeCurrency: this.offlineService.getHomeCurrency(),
+      homeCurrency: this.currencyService.getHomeCurrency(),
```
Julias0, line 245: getHomeCurrency is not cached. We should cache this

Julias0 (changes requested):
1 comment

Julias0 (approved):

PR #1375, opened 2022-09-23
FYLE-2v1r96m: Remove offline service methods -  Org User Settings

Dimple16:


src/app/core/services/offline.service.ts
```diff
@@ -63,35 +63,9 @@ export class OfflineService {
...
-          );
-        } else {
-          return from(this.storageService.get('cachedOrgUserSettings'));
-        }
-      })
-    );
-  }
-
   @Cacheable()
   getOrgUserMileageSettings() {
```
Julias0, line 67: We should remove off this method from offline service and move it to a different service

src/app/core/services/offline.service.ts
```diff
@@ -324,9 +298,9 @@ export class OfflineService {
 
   @Cacheable()
   getAllowedPaymentModes(): Observable<AccountType[]> {
```
Julias0, line 300: We should remove off this method from offline service and move it to a different service

Julias0 (changes requested):
Added comments

src/app/core/services/offline.service.ts
```diff
@@ -302,9 +271,9 @@ export class OfflineService {
...
-    return this.getOrgUserSettings().pipe(
-      map((orgUserSettings) => orgUserSettings?.payment_mode_settings?.allowed_payment_modes)
-    );
+    return this.orgUserSettingsService
```
Dimple16, line 274: removal of this method is taken care in another PR

Julias0 (approved):

PR #1387, opened 2022-09-28
FYLE-2v1r98f: Remove offline service methods - Categories

Dimple16:


src/app/core/services/offline.service.ts
```diff
@@ -104,30 +102,6 @@ export class OfflineService {
...
-          );
-        } else {
-          return from(this.storageService.get('cachedCategories'));
-        }
-      })
-    );
-  }
-
-  @Cacheable()
-  getAllEnabledCategories() {
```
Dimple16, line 125: this method is not required as we make the API call to get only active categories

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -1907,7 +1907,7 @@ export class AddEditExpensePage implements OnInit {
               recentValue: recentValues,
               recentCategories,
               etxn,
-              category: categories,
+              category: categories[0],
```
yksolanki9, line 1910: Add a check to see if category is not null. Else it'll throw error

yksolanki9 (approved):
Minor comment

Julias0 (approved):

PR #1390, opened 2022-09-28
FYLE-2v1r9cf: Remove offline service methods - Current Org

Dimple16:


src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -111,7 +111,7 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
         }
       });
     }
-    this.activeOrg$ = this.offlineService.getCurrentOrg();
+    this.activeOrg$ = this.orgService.getCurrentOrg();
```
yksolanki9, line 114: Make sure these methods are cached in org service

yksolanki9 (approved):

src/app/shared/components/sidemenu/sidemenu.component.ts
```diff
@@ -134,7 +136,7 @@ export class SidemenuComponent implements OnInit {
                 org_id: eou.ou.org_id,
                 org_user_id: eou.ou.id,
                 org_currency: currentOrg?.currency,
-                org_created_at: currentOrg?.created_at,
+                org_created_at: (currentOrg?.created_at).toString(),
```
Julias0, line 139: `currentOrg?.created_at?.toString()`

We should use optional chaining when using `.toString()`

Julias0 (changes requested):
1 comment

Julias0 (approved):

PR #1394, opened 2022-09-28
FYLE-2v1r9fv: Remove offline service methods - Current User

Dimple16:


src/app/shared/components/sidemenu/sidemenu.component.ts
```diff
@@ -93,7 +93,7 @@ export class SidemenuComponent implements OnInit {
         allowedActions: allowedActions$,
         deviceInfo: deviceInfo$,
         isSwitchedToDelegator: isSwitchedToDelegator$,
-        eou: this.offlineService.getCurrentUser(),
+        eou: this.orgUserService.getCurrent(),
```
yksolanki9, line 96: We already have a duplicate method `getEou()`  in `auth-service`. As we're anyways cleaning it, we should remove either of the two methods, and keep only one of them.

yksolanki9 (changes requested):

Julias0 (approved):

src/app/shared/components/sidemenu/sidemenu.component.ts
Dimple16: these are 2 different methods, 
in auth service, the method returns the `eou` from local storage of type promise, the use case here is different

yksolanki9 (approved):

PR #1396, opened 2022-09-30
Add session recording support with Smartlook

yksolanki9:
- Session recording will happen only for ICP users
- We'll start the recording once the user is on dashboard
- Will raise separate PR for random sampling of sign in page recording


src/app/core/services/smartlook.service.ts
```diff
@@ -0,0 +1,51 @@
...
+export class SmartlookService {
+  constructor(
+    private currencyService: CurrencyService,
+    private authService: AuthService,
+    private deviceService: DeviceService,
+    private smartlook: Smartlook
+  ) {}
+
+  init() {
+    forkJoin({
```
Julias0, line 25: This should only trigger if the user is connected to the internet. Add a check for isConnected

src/app/fyle/dashboard/dashboard.page.ts
```diff
@@ -86,6 +88,7 @@ export class DashboardPage implements OnInit {
 
   ionViewWillEnter() {
     this.setupNetworkWatcher();
+    this.smartlookService.init();
```
Julias0, line 91: For users who are loggin in, this will not work right?
We also need to set up in switch org. Lets add a check inside the service called isInitialized to refer if the smartlookService is live or not and based on that we can trigger session recordings

src/app/core/services/smartlook.service.ts
```diff
@@ -0,0 +1,51 @@
...
+      homeCurrency: this.currencyService.getHomeCurrency(),
+      eou: this.authService.getEou(),
+      deviceInfo: this.deviceService.getDeviceInfo(),
+    }).subscribe(({ homeCurrency, eou, deviceInfo }) => {
+      if (homeCurrency === 'USD') {
```
Julias0, line 30: if (homeCurrency === 'USD' && eou) {

src/app/core/services/smartlook.service.ts
```diff
@@ -0,0 +1,51 @@
...
+    private currencyService: CurrencyService,
+    private authService: AuthService,
+    private deviceService: DeviceService,
+    private smartlook: Smartlook
+  ) {}
+
+  init() {
+    forkJoin({
+      homeCurrency: this.currencyService.getHomeCurrency(),
+      eou: this.authService.getEou(),
```
Julias0, line 27: getEou returns a promise. Will that work properly with ForkJoin? Shouldnt we wrap it inside a `from`?

Julias0 (changes requested):
Some comments

src/app/core/services/smartlook.service.ts
yksolanki9: forkJoin converts the promise into observable, will still add wrap inside `from`

src/app/fyle/dashboard/dashboard.page.ts
yksolanki9: Ignoring this comment as discussed

src/app/core/services/smartlook.service.ts
yksolanki9: Added this in filter operator

src/app/core/services/smartlook.service.ts
yksolanki9: Done

Julias0 (approved):

PR #1398, opened 2022-10-03
#mission_nps_mobile_40: For sent back reports, show comments section when opening the report

yksolanki9:


src/app/fyle/my-view-report/my-view-report.page.ts
```diff
@@ -238,6 +240,11 @@ export class MyViewReportPage implements OnInit {
       this.reportCurrencySymbol = getCurrencySymbol(erpt?.rp_currency, 'wide');
       this.reportName = erpt?.rp_purpose;
       this.reportId = erpt?.rp_id;
+
+      //For sent back reports, show the comments section instead of expenses when opening the report
+      if (erpt.rp_state === 'APPROVER_INQUIRY') {
```
Dimple16, line 245: can you check the `segmentChanged()` once, just to verify if it'll overwrite this value

Dimple16: we should be handling in one way itself, here we have a separate value in HTML, and the we are setting `ionsegment` value here,

Dimple16 (changes requested):

Dimple16 (approved):

src/app/fyle/my-view-report/my-view-report.page.ts
yksolanki9: Discussed on slack

PR #1410, opened 2022-10-11
Task Filter Bug Fix

jayfyle:
## Task : Change the name of the filter for tasks

The name of the filter was given as Draft which has been deprecated, changed it to Incomplete in this fix

image


src/app/core/services/tasks.service.ts
```diff
@@ -19,6 +19,7 @@ import { UserEventService } from './user-event.service';
 import { HandleDuplicatesService } from './handle-duplicates.service';
 import { DuplicateSet } from '../models/v2/duplicate-sets.model';
 import { CurrencyService } from './currency.service';
+import { filter } from 'lodash';
```
Julias0, line 22: Why is this required?

Julias0 (changes requested):

Julias0 (approved):

PR #1415, opened 2022-10-12
Add new var for tracking if camera preview has been initiated

yksolanki9:
Issue:
- We're getting errors like `camera already started` and `camera already stopped` very frequently
- Previously, we used to track camera preview using a flag called `isCameraShown`, but the problem here is that we used to set this flag only when the camera started or stopped, and not when we initiated the request to start/stop the camera
- The camera plugin takes some time to start/stop the camera preview, and hence we may get the above errors if the user tries to open the camera in between
- Also, we cannot change `isCameraShown` to track when the camera start/stop request was initiated, as we use it hide the parent components, and if we do so, the the user would see a weird UI till the camera preview actually starts

Solution
- Created a new flag `isCameraPreviewInitiated` to track whether the user has intitiated a camera preview - if yes, we wont intitiated it again. Same goes for stopping the preview
- The existing flag `isCameraShown` is renamed to `isCameraPreviewStarted` which aptly signifies the purpose of this flag.

Sentry bugs:
- [Uncaught (in promise): Error: camera already started](https://sentry.io/organizations/fyle-technologies-private-limi/issues/3356869620/events/d6172a8d118b43859bc8769450c5e938/?project=5622998&statsPeriod=14d)
- [Uncaught (in promise): Error: camera already stopped](https://sentry.io/organizations/fyle-technologies-private-limi/issues/3552152935/events/dc3d88e2a9f240ed802d68771dc88089/?project=5622998&statsPeriod=14d)

### DIFF
--- a/src/app/fyle/add-edit-advance-request/add-edit-advance-request.page.html
+++ b/src/app/fyle/add-edit-advance-request/add-edit-advance-request.page.html
@@ -1,4 +1,4 @@
-<ion-header mode="md" [ngClass]="{'d-none': isCameraShown}">
+<ion-header mode="md" [ngClass]="{'d-none': isCameraPreviewStarted}">
   <ion-toolbar mode="md" class="add-edit-advance-request--toolbar">
     <ion-buttons
       mode="md"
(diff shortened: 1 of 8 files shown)


Dimple16 (approved):

Dimple16:
looks good, lets test this on multiple devices,

Julias0 (approved):

PR #1419, opened 2022-10-14
FYLE-2uw0r29: Fix for cards of different org showing up on dashboard and clean up

Dimple16:
### What was the issue?
- There were 2 variables used to get the cardTransactionsAndDetails
- One was the observable `cardTransactionsAndDetails$` and the other was the subscribed list value `cardTransactionsAndDetails`
- When the corporate card settings were enabled, cardTransactionsAndDetails$ and cardTransactionsAndDetails both got set
- In the else block, only the observable was set to null, the `cardTransactionsAndDetails` list was not set to an empty list
- Due to this, when we switched to an org with the settings disabled from an org that had card details, it showed up the previous data, basically, the list didn't get reset at all
- This also happened because we didn't call the initialize method if the settings were disabled, that's why the value never got reset

### What is the fix?
- I have added the change to reset the `cardTransactionsAndDetails` list to an empty list

### Refactoring and cleaning up
- There were a few pieces of code that were lying around even though it didn't matter now related to the `#unify_ccc_expenses_flow` initiative, these were supposed to be removed during `#fix_implicit_merge changes`
- I have removed a bunch of code and cleaned up a few unnecessary conditions

### DIFF
--- a/src/app/fyle/dashboard/stats/stats.component.html
+++ b/src/app/fyle/dashboard/stats/stats.component.html
@@ -95,72 +95,50 @@
     </ion-grid>
   </div>
 
-  <ng-container *ngIf="isConnected$ | async; else offlineMessage"></ng-container>
-
-  <div *ngIf="(isConnected$ | async) && cardTransactionsAndDetails?.length > 0" class="stats--ccc-cards-section">
-    <div>
-      <ion-grid
-        [ngClass]="{
-          'stats--ccc-cards-grid': !isUnifyCCCExpensesSettings,
-          'stats--ccc-cards-grid-multiple': isUnifyCCCExpensesSettings
-        }"
-        class=""
-      >
-        <ion-row>
-          <ion-col>
-            <div class="stats--header stats--ccc-cards-header">CARD SPEND</div>
-          </ion-col>
-        </ion-row>
-      </ion-grid>
-      <ng-container *ngIf="isUnifyCCCExpensesSettings">
-        <app-spent-cards
-          *ngIf="cardTransactionsAndDetails?.length > 1"
-          [spentCards]="cardTransactionsAndDetails"
-          [homeCurrency]="homeCurrency$"
-          [currencySymbol]="currencySymbol$"
-        ></app-spent-cards>
-        <div *ngIf="cardTransactionsAndDetails?.length === 1" class="stats--single-ccc">
-          <app-card-detail
-            [cardDetail]="cardTransactionsAndDetails[0]"
+  <ng-container *ngIf="isConnected$ | async; else offlineMessage">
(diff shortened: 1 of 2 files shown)


src/app/fyle/dashboard/stats/stats.component.html
```diff
@@ -132,77 +129,6 @@
         </div>
       </ng-container>
 
-      <ng-container *ngIf="!isUnifyCCCExpensesSettings">
```
Dimple16, line 135: this section is not required since there is no more `nonunifyccc`, all the orgs have it enabled by default

src/app/fyle/dashboard/stats/stats.component.ts
```diff
@@ -136,45 +130,13 @@ export class StatsComponent implements OnInit {
...
-        shareReplay(1)
-      );
-      this.cardTransactionsAndDetailsNonUnifyCCC$.subscribe((details) => {
-        this.cardTransactionsAndDetailsNonUnifyCCC = details;
+    this.dashboardService
+      .getCCCDetails()
+      .pipe()
+      .subscribe((details) => {
+        this.cardTransactionsAndDetails = this.getCardDetail(details.cardDetails);
         this.isCCCStatsLoading = false;
```
Julias0, line 138: put this in finalize block

src/app/fyle/dashboard/stats/stats.component.html
```diff
@@ -97,10 +97,7 @@
 
   <ng-container *ngIf="isConnected$ | async; else offlineMessage"></ng-container>
 
-  <div
-    *ngIf="(isConnected$ | async) && (cardTransactionsAndDetails?.length > 0 || cardTransactionsAndDetailsNonUnifyCCC)"
-    class="stats--ccc-cards-section"
-  >
+  <div *ngIf="(isConnected$ | async) && cardTransactionsAndDetails?.length > 0" class="stats--ccc-cards-section">
```
Julias0, line 100: Lets move the isConnected to a parent <ng-container>

src/app/fyle/dashboard/stats/stats.component.html
```diff
@@ -95,197 +95,123 @@
     </ion-grid>
   </div>
 
-  <ng-container *ngIf="isConnected$ | async; else offlineMessage"></ng-container>
```
Dimple16, line 98: change -> added the below section inside the parent container, no other change

yksolanki9 (changes requested):
Can you add some notes on what was the issue here and how is it fixed? Its difficult to understand from the changes.
Also, please fix lint errors

Julias0 (approved):

PR #1425, opened 2022-10-17
Draft Button Bug Fix

jayfyle:
## Report Title Draft Button Fix

The draft button would be active and allow to create a report even if the report title is empty

### DIFF
--- a/src/app/fyle/my-create-report/my-create-report.page.ts
+++ b/src/app/fyle/my-create-report/my-create-report.page.ts
@@ -68,7 +68,7 @@ export class MyCreateReportPage implements OnInit {
   ) {}
 
   detectTitleChange() {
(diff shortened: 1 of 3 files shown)


src/app/fyle/my-create-report/my-create-report.page.ts
```diff
@@ -65,6 +67,17 @@ export class MyCreateReportPage implements OnInit {
     private refinerService: RefinerService
   ) {}
 
+  detectTitleChange() {
+    if (this.reportTitle === '') {
```
Julias0, line 71: `if (this.reportTitle && this.reportTitle === '') {`
This is a failsafe just in case this gets set to null or undefined

Julias0 (approved):

PR #1426, opened 2022-10-17
Fyle 2wuxjb5

Aditya-shrivastav:
Resend Invite Button added and change content for warning message

https://user-images.githubusercontent.com/56168371/196190913-4b41cad2-0b91-42c3-a272-47eae876bd68.mov


src/app/core/services/auth.service.ts
```diff
@@ -57,8 +58,8 @@ export class AuthService {
     );
   }
 
-  resendVerification(email: string) {
-    return this.apiService.post('/auth/resend_email_verification', { email });
+  resendVerification(email: string, org_id?: string): Observable<ResendEmailVerificationResponse> {
```
abhishek1234321, line 61: `orgId`

abhishek1234321: why is this optional here?

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -146,13 +152,69 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
...
+          this.signOut();
+        }
+      });
+    } else if (action === 'resend') {
+      // If user clicks on resend Button, Resend Invite to the user and then logout if user have only one org.
+      this.resendInvite(email, orgId)
+        .pipe(
+          switchMap(() => this.orgs$),
+          catchError((error) => {
+            this.showToastNotification('Oops! Something went wrong. Please try again.');
```
hlkavya0213, line 189: change the content get it from Bhavika

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -146,13 +152,69 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
...
+            return throwError(() => error);
+          })
+        )
+        .subscribe((orgs) => {
+          if (orgs.length === 1) {
+            this.signOut();
+          }
+          this.showToastNotification('Verification Email Sent');
```
hlkavya0213, line 197: get content from Bhavika

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -163,18 +225,7 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
...
-       * After showing the alert, the user will be redirected to the sign-in page since there is no other org they are a part of.
-       * If the user has more than 1 org, the user will stay on the switch org page to choose another org.
-       */
-      this.orgs$.subscribe((orgs) => {
-        if (orgs.length === 1) {
-          this.signOut();
-        }
-      });
-    }
+    this.handlePopupDismiss(data.action, email, orgId);
```
Dimple16, line 228: `handleDismissPopup`? or mention the dismiss action name

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -146,13 +152,69 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
...
+
+    this.matSnackBar.openFromComponent(ToastMessageComponent, {
+      ...this.snackbarProperties.setSnackbarProperties('success', toastMessageData),
+      panelClass: ['msb-info'],
+    });
+    this.trackingService.showToastMessage({ ToastContent: toastMessageData.message });
+  }
+
+  handlePopupDismiss(action: string, email: string, orgId: string) {
+    if (action === 'close') {
```
Dimple16, line 172: this block doesn't return/do anything if the `orgs.length` condition fails, is this expected?

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -146,13 +152,69 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
...
+       * Case: When a user is added to an SSO org but hasn't verified their account through the link.
+       * After showing the alert, the user will be redirected to the sign-in page since there is no other org they are a part of.
+       * If the user has more than 1 org, the user will stay on the switch org page to choose another org.
+       */
+      this.orgs$.subscribe((orgs) => {
+        if (orgs.length === 1) {
+          this.signOut();
+        }
+      });
+    } else if (action === 'resend') {
```
Dimple16, line 183: will there be a case of hitting an `else` block here? how do we handle if the action is null/undefined? test this and add fail-safe if required

src/app/core/services/auth.service.ts
```diff
@@ -57,8 +58,8 @@ export class AuthService {
     );
   }
 
-  resendVerification(email: string) {
-    return this.apiService.post('/auth/resend_email_verification', { email });
+  resendVerification(email: string, orgId?: string): Observable<ResendEmailVerificationResponse> {
```
Dimple16, line 61: `resendEmailVerification`

Dimple16 (changes requested):

src/app/auth/switch-org/switch-org.page.ts
Aditya-shrivastav: Yes, this is fine. We just need to check if user have only 1 org. If yes, we will logout him. Otherwise, we will do nothing.

src/app/auth/switch-org/switch-org.page.ts
Aditya-shrivastav: else block is not required, since actions are defined inside popup itself. Although, changed condition to have a fail-safe

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -146,24 +152,40 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
...
+      // If user clicks on resend Button, Resend Invite to the user and then logout if user have only one org.
+      this.resendInvite(email, orgId)
+        .pipe(
+          switchMap(() => this.orgs$),
+          catchError((error) => {
+            this.showToastNotification('Verification link could not be sent. Please try again!');
+            return throwError(() => error);
+          })
+        )
+        .subscribe((orgs) => {
```
Dimple16, line 182: why add `this.orgs$` inside the `switchMap` here? can directly use it below and subscribe, same as the else block

Dimple16 (changes requested):

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -146,37 +152,84 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
...
+    this.matSnackBar.openFromComponent(ToastMessageComponent, {
+      ...this.snackbarProperties.setSnackbarProperties('success', toastMessageData),
+      panelClass: ['msb-info'],
     });
-    await popover.present();
+    this.trackingService.showToastMessage({ ToastContent: toastMessageData.message });
+  }
 
-    const { data } = await popover.onWillDismiss();
+  logoutIfSingleOrg() {
```
Dimple16, line 171: `logoutForSingleOrg`?

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -146,37 +152,84 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
...
+    /*
+     * Case: When a user is added to an SSO org but hasn't verified their account through the link.
+     * After showing the alert, the user will be redirected to the sign-in page since there is no other org they are a part of.
+     * If the user has more than 1 org, the user will stay on the switch org page to choose another org.
+     */
+    this.orgs$.subscribe((orgs) => {
```
Dimple16, line 177: you don't have to subscribe to `this.orgs$` here, 
check `ionViewWillEnter()` in line 97, we subscribe it once, and you can assign `this.orgs` there, use it here directly

Dimple16 (changes requested):

Dimple16 (approved):

PR #1438, opened 2022-10-25
Functions to accomodate check policies call

KirtiGautam:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=7484Ucj)

### DIFF
--- a/src/app/core/models/platform/platform-check-expense-policy.model.ts
+++ b/src/app/core/models/platform/platform-check-expense-policy.model.ts
@@ -1,0 +1,9 @@
+import { FinalExpensePolicyState } from './platform-final-expense-policy-state.model';
+import { IndividualExpensePolicyState } from './platform-individual-expense-policy-state.model';
+
+export interface CheckExpensePolicies {
+  data: {
+    individual_desired_states: IndividualExpensePolicyState[];
+    final_desired_state: FinalExpensePolicyState;
+  };
+}
--- a/src/app/core/models/platform/platform-final-expense-policy-state.model.ts
+++ b/src/app/core/models/platform/platform-final-expense-policy-state.model.ts
@@ -1,7 +1,6 @@
 export interface FinalExpensePolicyState {
   add_approver_user_ids: string[];
-  // Expense Id is absent in check policies call
-  expense_id?: string;
+  expense_id?: string; // Expense Id is absent in check policies call
   amount: number;
   flag: boolean;
   is_receipt_mandatory: boolean;
--- a/src/app/core/models/platform/platform-final-expense-policy-state.model.ts
+++ b/src/app/core/models/platform/platform-final-expense-policy-state.model.ts
@@ -1,6 +1,7 @@
 export interface FinalExpensePolicyState {
   add_approver_user_ids: string[];
-  expense_id: string;
+  // Expense Id is absent in check policies call
+  expense_id?: string;
   amount: number;
   flag: boolean;
   is_receipt_mandatory: boolean;
--- a/src/app/core/models/platform/platform-individual-expense-policy-state.model.ts
+++ b/src/app/core/models/platform/platform-individual-expense-policy-state.model.ts
@@ -7,12 +7,12 @@ export interface IndividualExpensePolicyState {
   // In check policies call, the object key is returned with different name than in expense_policies_states call
   expense_policy_rule?: {
     id: string;
-    description: number;
+    description: string;
     action_show_warning: boolean;
   };
   rule?: {
     id: string;
-    description: number;
+    description: string;
     action_show_warning: boolean;
   };
   amount: number;
--- a/src/app/core/models/platform/platform-individual-expense-policy-state.model.ts
+++ b/src/app/core/models/platform/platform-individual-expense-policy-state.model.ts
@@ -2,9 +2,15 @@ export interface IndividualExpensePolicyState {
   add_approver_user_ids: string[];
   run_status: string;
   run_result: string[];
-  expense_id: string;
+  expense_id?: string; // This key is absent in case of check policies call
   expense_policy_rule_id: string;
-  rule: {
+  // In check policies call, the object key is returned with different name than in expense_policies_states call
+  expense_policy_rule?: {
+    id: string;
+    description: number;
+    action_show_warning: boolean;
+  };
+  rule?: {
     id: string;
     description: number;
     action_show_warning: boolean;
--- a/src/app/core/services/policy.service.ts
+++ b/src/app/core/services/policy.service.ts
@@ -1,6 +1,7 @@
 import { Injectable } from '@angular/core';
 import { map, Observable } from 'rxjs';
 import { PlatformApiResponse } from '../models/platform/platform-api-response.model';
+import { CheckExpensePolicies } from '../models/platform/platform-check-expense-policy.model';
 import { ExpensePolicyStates } from '../models/platform/platform-expense-policy-states.model';
 import { IndividualExpensePolicyState } from '../models/platform/platform-individual-expense-policy-state.model';
 import { PolicyViolation } from '../models/policy-violation.model';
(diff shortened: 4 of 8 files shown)


src/app/core/models/platform/platform-check-expense-policy.model.ts
```diff
@@ -0,0 +1,9 @@
+import { FinalExpensePolicyState } from './platform-final-expense-policy-state.model';
+import { IndividualExpensePolicyState } from './platform-individual-expense-policy-state.model';
+
+export interface CheckExpensePolicies {
```
Dimple16, line 4: why name it `checkExpensePolicies`? 
can we name it something like `ExpensePolicy`? since this isn't a list

src/app/core/services/policy.service.ts
```diff
@@ -16,24 +17,24 @@ export class PolicyService {
     private approverPlatformApiService: ApproverPlatformApiService
   ) {}
 
-  getCriticalPolicyRules(result) {
+  getCriticalPolicyRules(result: CheckExpensePolicies): string[] {
```
Dimple16, line 20: `result` rename with context

src/app/core/models/platform/platform-final-expense-policy-state.model.ts
```diff
@@ -1,6 +1,7 @@
 export interface FinalExpensePolicyState {
   add_approver_user_ids: string[];
-  expense_id: string;
+  // Expense Id is absent in check policies call
```
Dimple16, line 3: add inline comment itself, same like below, easier to read

src/app/core/models/platform/platform-individual-expense-policy-state.model.ts
```diff
@@ -2,9 +2,15 @@ export interface IndividualExpensePolicyState {
   add_approver_user_ids: string[];
   run_status: string;
   run_result: string[];
-  expense_id: string;
+  expense_id?: string; // This key is absent in case of check policies call
   expense_policy_rule_id: string;
-  rule: {
+  // In check policies call, the object key is returned with different name than in expense_policies_states call
```
Dimple16, line 7: so we will have 2 things with the same values, but different names, is this intentional in the backend?

src/app/core/services/policy.service.ts
```diff
@@ -16,24 +17,24 @@ export class PolicyService {
...
     const criticalPopupRules = [];
 
-    result.transaction_policy_rule_desired_states.forEach((desiredState) => {
-      if (typeof desiredState.policy_amount === 'number' && desiredState.policy_amount < 0.0001) {
-        criticalPopupRules.push(desiredState.description);
+    result.data.individual_desired_states.forEach((desiredState) => {
```
Dimple16, line 23: if we change this here, we can't push this change until all the versions use the new API right? else it'll break

src/app/core/services/policy.service.ts
```diff
@@ -16,24 +17,24 @@ export class PolicyService {
...
+      if (typeof desiredState.amount === 'number' && desiredState.amount < 0.0001) {
+        criticalPopupRules.push(desiredState.expense_policy_rule.description);
       }
     });
 
     return criticalPopupRules;
   }
 
-  getPolicyRules(result) {
+  getPolicyRules(result: CheckExpensePolicies): string[] {
```
Dimple16, line 32: rename `result` with context

src/app/core/services/transaction.service.ts
```diff
@@ -484,6 +488,50 @@ export class TransactionService {
     return data;
   }
 
+  checkPolicy(policyExpense: PolicyExpense): Observable<CheckExpensePolicies> {
```
Dimple16, line 491: `CheckExpensePolicies` this model will be renamed

src/app/core/services/transaction.service.ts
```diff
@@ -484,6 +488,50 @@ export class TransactionService {
...
+    return this.orgUserSettingsService.get().pipe(
+      switchMap((orgUserSettings) => {
+        // setting txn_dt time to T10:00:00:000 in local time zone
+        if (policyExpense.spent_at) {
```
Dimple16, line 495: will `policyExpense` always have a proper value? else we'll have to add optional chaining

src/app/shared/components/fy-policy-violation/fy-policy-violation.component.html
```diff
@@ -29,31 +29,31 @@
         <app-policy-violation-action
           [icon]="'../../../../assets/svg/flag-outline.svg'"
           [message]="'Expense will be flagged for verification and approval'"
-          [actionsCount]="availableActionsCount"
+          [actionsCount]="policyAction?.run_summary?.length"
```
Dimple16, line 32: will this ever be an empty list with length 0? 
If this is the case, we might need to check the component logic if it handles both cases

src/app/shared/components/fy-policy-violation/fy-policy-violation.component.ts
```diff
@@ -59,7 +63,14 @@ export class FyPolicyViolationComponent implements OnInit {
 
   constructCappingAction() {
     if (this.isExpenseCapped) {
-      const cappedAmountMatches = this.utilityService.getAmountWithCurrencyFromString(this.policyActionDescription);
+      console.log('here in capping');
```
Dimple16, line 66: remove logs

Dimple16 (changes requested):

src/app/core/models/platform/platform-individual-expense-policy-state.model.ts
KirtiGautam: Hey, no either of these would be present.
In check_policies, it would be returned as `expense_policy_states` and in previous call it was returned as `rule`

src/app/core/services/policy.service.ts
KirtiGautam: Right now, the policy is running on both platform and the public side. This change can go out and people will slowly move from public to platform. After all clients are migrated, then only we will kill the public backend service

src/app/core/services/transaction.service.ts
KirtiGautam: `spent_at` is always needed. So, it will always be present

src/app/shared/components/fy-policy-violation/fy-policy-violation.component.html
KirtiGautam: Yes, this can be zero. But this variable was initialised with zero previously so I don't think it would break now.

Dimple16: okay, make sure we have a default value set for all the variables OR add fail-safe logic

src/app/shared/components/fy-policy-violation/fy-policy-violation.component.ts
```diff
@@ -75,23 +85,11 @@ export class FyPolicyViolationComponent implements OnInit {
...
-      const fns = [
-        this.policyService.isExpenseFlagged,
-        this.policyService.isPrimaryApproverSkipped,
-        this.policyService.needAdditionalApproval,
-        this.policyService.isExpenseCapped,
-      ];
-
-      const statuses = fns.map((fn) => fn(this.policyActionDescription));
-      const [isExpenseFlagged, isPrimaryApproverSkipped, needAdditionalApproval, isExpenseCapped] = statuses;
-      this.availableActionsCount = statuses.reduce((acc, curr) => (curr ? acc + 1 : acc), 0);
```
Dimple16, line 94: let's make sure to remove all the instances of `availableActionsCount` /  `statuses` if they aren't used anymore across

src/app/core/services/transaction.service.ts
```diff
@@ -484,6 +488,50 @@ export class TransactionService {
...
+        }
+        const payload = {
+          data: platformPolicyExpense,
+        };
+        return this.spenderPlatformApiService.post<ExpensePolicy>('/expenses/check_policies', payload);
+      })
+    );
+  }
+
   testPolicy(etxn) {
```
Dimple16, line 535: what about this method? since we have the previous one, shouldn't we remove it if unused?

Dimple16 (changes requested):
some clarifications

src/app/shared/components/fy-policy-violation/fy-policy-violation.component.ts
KirtiGautam: Yes, it's not used anywhere so have removed it from the file.

Dimple16 (approved):

PR #1440, opened 2022-10-25
Fix for project<>category mapping issues

Dimple16:
### What was the issue?
- When the project got preset, the project watcher did not trigger, due to which the allowed categories filter method did not get called

### What is the fix?
- I have explicitly added checks to filter the allowed categories for recent categories as well as the all-category list
- As soon as the project is preset, we filter out the allowed categories
- We will autofill category only if it is allowed for that particular project
- Setting the form group project value explicitly before the patch which triggers the project watcher and filters out the allowed categories

### DIFF
--- a/src/app/fyle/add-edit-expense/add-edit-expense.page.ts
+++ b/src/app/fyle/add-edit-expense/add-edit-expense.page.ts
@@ -349,8 +349,6 @@ export class AddEditExpensePage implements OnInit {
 
   breakfastSystemCategories: string[];
 
-  canAutofillCategory: boolean;
-
   constructor(
     private activatedRoute: ActivatedRoute,
     private accountsService: AccountsService,
@@ -1643,6 +1641,7 @@ export class AddEditExpensePage implements OnInit {
 
           this.recentCurrencies = recentCurrencies;
 
+          let canAutofillCategory: boolean;
           /* Autofill project during these cases:
            * 1. Autofills is allowed and enabled
            * 2. During add expense - When project field is empty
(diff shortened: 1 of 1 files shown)


src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -1665,9 +1658,36 @@ export class AddEditExpensePage implements OnInit {
...
               this.presetProjectId = project.project_id;
+
+              // Check if the recent categories are allowed for the project auto-filled
+              const isAllowedRecentCategories = recentCategories.map((category) =>
+                project.project_org_category_ids.indexOf(category.value.id) > -1 ? true : false
+              );
+
+              // Set the updated allowed recent categories
+              this.recentCategories = recentCategories.filter(
+                (category) => project.project_org_category_ids.indexOf(category.value.id) > -1
```
Julias0, line 1669: same as above

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -1665,9 +1658,36 @@ export class AddEditExpensePage implements OnInit {
...
+              );
+
+              // Only if the most recent category is allowed for the auto-filled project, category field can be auto-filled
+              this.canAutofillCategory = isAllowedRecentCategories[0];
```
Julias0, line 1673: do you need this to be connected to `this`? You are using this locally inside the function only

Julias0 (changes requested):
Added comments

yksolanki9 (approved):

Julias0 (approved):

PR #1446, opened 2022-10-26
Replace /policy/test with /check_policies in split expense page

KirtiGautam:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=4zwZp8I)


src/app/core/services/split-expense.service.ts
```diff
@@ -195,10 +206,14 @@ export class SplitExpenseService {
 
   runPolicyCheck(etxns: Expense[], fileObjs: FileObject[]): Observable<{ [transactionID: string]: PolicyViolation }> {
     if (etxns?.length > 0) {
+      const platformExpenses = [];
       etxns.forEach((etxn) => {
-        etxn.tx_num_files = fileObjs ? fileObjs.length : 0;
+        // transformTo method requires unflattend transaction object
+        const platformExpense = this.dataTransformService.unflatten(etxn).tx;
```
Dimple16, line 212: can we use a different name here? both `platformExpenses` and `platformExpense` used together is confusing

Dimple16: you can call it `platformExpensesList`

Dimple16 (changes requested):

Dimple16 (approved):

PR #1450, opened 2022-10-27
Default select of report

shreya-fyle1:


src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -1380,22 +1380,26 @@ export class AddEditExpensePage implements OnInit {
...
-          );
+          return reportOptions
+            .map((res) => res.value)
+            .find((reportOption) => reportOption.rp.id === this.activatedRoute.snapshot.params.rp_id);
+        } else if (!autoSubmissionReportName) {
+          return reportOptions.map((res: any) => {
+            if (reportOptions.length === 1 && reportOptions[0].value.rp.state === 'DRAFT') {
+              return reportOptions[0].value;
+            } else {
+              this.addToNewReport(res.id);
```
yksolanki9, line 1400: Why are we creating a new report here?

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -1380,22 +1380,26 @@ export class AddEditExpensePage implements OnInit {
...
-              reportOptions
-                .map((res) => res.value)
-                .find((reportOption) => reportOption.rp.id === this.activatedRoute.snapshot.params.rp_id)
-            )
-          );
+          return reportOptions
+            .map((res) => res.value)
+            .find((reportOption) => reportOption.rp.id === this.activatedRoute.snapshot.params.rp_id);
+        } else if (!autoSubmissionReportName) {
+          return reportOptions.map((res: any) => {
```
yksolanki9, line 1396: Why is this map needed?

yksolanki9 (changes requested):

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
Dimple16: this isn't required here anymore, since `reportOptions` is already resolved @shreya-fyle1

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
Dimple16: this shouldn't be added here, we are just setting a default value, not creating another report

Dimple16 (changes requested):

Dimple16 (approved):

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -1380,22 +1380,26 @@ export class AddEditExpensePage implements OnInit {
...
-              reportOptions
-                .map((res) => res.value)
-                .find((reportOption) => reportOption.rp.id === this.activatedRoute.snapshot.params.rp_id)
-            )
-          );
+          return reportOptions
+            .map((res) => res.value)
+            .find((reportOption) => reportOption.rp.id === this.activatedRoute.snapshot.params.rp_id);
+        } else if (!autoSubmissionReportName) {
+          return reportOptions.map((res) => {
```
Dimple16, line 1396: since we return the report value/null in a certain case and an observable by default, this map is required

yksolanki9 (changes requested):
Discussed on slack

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -1386,22 +1386,24 @@ export class AddEditExpensePage implements OnInit {
...
-          );
+          return reportOptions
+            .map((res) => res.value)
+            .find((reportOption) => reportOption.rp.id === this.activatedRoute.snapshot.params.rp_id);
+        } else if (
+          !autoSubmissionReportName &&
+          reportOptions.length === 1 &&
+          reportOptions[0].value.rp.state === 'DRAFT'
+        ) {
+          return of(reportOptions[0].value);
```
yksolanki9, line 1406: We're returning an observable here, but returning a value in the first two conditions. The first two conditions would break here.
You can return observables there, but the better solution is to use `map` here and return the values directly

yksolanki9 (changes requested):

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -1,3 +1,4 @@
+import { Value } from './../../core/models/v2/stats-response.model';
```
yksolanki9, line 1: Remove this

yksolanki9 (approved):
Remove the unnecessary import and test all the cases before merging

PR #1458, opened 2022-11-02
Report Redirect Fix

jayfyle:
## Redirection from Add Expense to Report

When creating an expense from the report, the app redirects to the expenses page instead of going to the report where the expense was created.

### Edit 1: The redirection works but happens before the expense has been added to the report

### DIFF
--- a/src/app/fyle/add-edit-expense/add-edit-expense.page.ts
+++ b/src/app/fyle/add-edit-expense/add-edit-expense.page.ts
@@ -352,6 +352,8 @@ export class AddEditExpensePage implements OnInit {
 
   breakfastSystemCategories: string[];
 
+  reportId: string;
+
   constructor(
     private activatedRoute: ActivatedRoute,
     private accountsService: AccountsService,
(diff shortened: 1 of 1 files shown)


src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -969,6 +971,10 @@ export class AddEditExpensePage implements OnInit {
   }
 
   ngOnInit() {
+    if (this.activatedRoute.snapshot.params.rp_id) {
+      this.reportId = this.activatedRoute.snapshot.params.rp_id;
```
Julias0, line 975: Instead of doing this you can create a function to get you this directly

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -3729,6 +3736,8 @@ export class AddEditExpensePage implements OnInit {
         this.saveAndNewExpenseLoader = false;
         this.saveAndNextExpenseLoader = false;
         this.saveAndPrevExpenseLoader = false;
+
+        this.router.navigate(['/', 'enterprise', 'my_view_report', { id: this.reportId }]);
```
Julias0, line 3740: We shouldnt do this in finalize

PR #1463, opened 2022-11-05
Removed unused currency settings

abhishek1234321:


src/app/core/services/org-settings.service.ts
```diff
@@ -246,30 +246,6 @@ export class OrgSettingsService {
         id: 'CURRENCYLAYER',
         name: 'Currency Layer',
       },
-      fixer_provider_settings: {
```
abhishek1234321, line 249: We will be deprecating this cc @jayachandhar

Julias0 (approved):

PR #1469, opened 2022-11-07
#mission_nps_mobile_40: Prompt user to use bulkfyle after using instafyle thrice

yksolanki9:
UI when camera opened for the 4th time:
https://user-images.githubusercontent.com/39493102/200512391-f8485e76-b742-46fd-bd0a-edcf33a49404.mp4

UI Preview:
Screenshot 2022-11-08 at 1 12 52 PM

### DIFF
--- a/src/app/core/services/transactions-outbox.service.ts
+++ b/src/app/core/services/transactions-outbox.service.ts
@@ -32,6 +32,9 @@ export class TransactionsOutboxService {
 
   ROOT_ENDPOINT: string;
 
+  //Used for showing bulk mode prompt when count > 3
+  singleCaptureCount = 0;
+
   constructor(
     private storageService: StorageService,
     private dateService: DateService,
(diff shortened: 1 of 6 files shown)


src/app/core/services/transactions-outbox.service.ts
```diff
@@ -32,6 +32,9 @@ export class TransactionsOutboxService {
 
   ROOT_ENDPOINT: string;
 
+  //Used for showing bulk mode prompt when instafyle is used more than thrice in the same session
+  singleCaptureCount = 0;
```
Julias0, line 36: make this variable private and allow access throw methods only. Giving variable access inside services is an antipattern

Julias0: getSingleCaptureCountInSameSession
setSingleCaptureCountInSameSession

Julias0: Also attach a video recording for me to review the interaction

Julias0 (changes requested):
Added comments

src/app/core/services/transactions-outbox.service.ts
yksolanki9: Have added a getter method and another method to increment the count.
Have also attached a recording - Sorry for this

Julias0 (approved):

PR #1471, opened 2022-11-08
#mission_nps_mobile_40: Fix redirection when expense is added to a report or a reported expense is edited

yksolanki9:
Fixes redirection in the following cases:
- When expense is created from expense form and report is selected in the add to report field
- When a new expense is created from inside report
- When an existing reported expense is edited

Also, moved some variable/observable initialization from `ngOnInit` to `ionViewWillEnter` inside view reports page, else if the page was pulled from stack, the report details were not updated.

Will raise separate PR for mileage/per-diem

### DIFF
--- a/src/app/fyle/add-edit-expense/add-edit-expense.page.ts
+++ b/src/app/fyle/add-edit-expense/add-edit-expense.page.ts
@@ -3062,7 +3062,12 @@ export class AddEditExpensePage implements OnInit {
 
               that
                 .addExpense('SAVE_EXPENSE')
-                .pipe(switchMap((txnData: Promise<any>) => from(txnData)))
+                .pipe(
+                  switchMap((txnData: Promise<any>) => from(txnData)),
+                  finalize(() => {
+                    this.saveExpenseLoader = false;
+                  })
+                )
                 .subscribe((res: any) => {
                   if (that.fg.controls.add_to_new_report.value && res?.transaction) {
                     this.addToNewReport(res.transaction.id);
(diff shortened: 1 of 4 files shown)


src/app/fyle/my-view-report/my-view-report.page.ts
```diff
@@ -151,7 +151,38 @@ export class MyViewReportPage implements OnInit {
     });
   }
 
-  ngOnInit() {
```
yksolanki9, line 154: Report details were not getting updated when the place was already in stack. So moved the variable initialization to `ionViewWillEnter()`

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -3732,7 +3738,6 @@ export class AddEditExpensePage implements OnInit {
         )
       ),
       finalize(() => {
-        this.saveExpenseLoader = false;
```
Julias0, line 3735: why remove this?
What happens if the addEntryAndSync call fails?

Julias0 (changes requested):
Added minor comments

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
yksolanki9: @Julias0 have not removed it, moved to line 3068 so that the loader stops only after we have the promise is resolved and we have the txn object

Julias0 (approved):

PR #1482, opened 2022-11-10
#mission_nps_mobile_40: Fix hardware back button redirections on tasks page

yksolanki9:
Changes:
- When navigating to the tasks page from dashboard, you'd be redirected back to the dashboard
- When navigating to tasks page from any other page, you'd be redirected to that page|
- Removed all tasks filters when user goes to tasks page from personal cards or team advances page
- Fixed a bug where clicking on hw back button on expenses page did not do anything

### DIFF
--- a/src/app/core/services/shared.service.ts
+++ b/src/app/core/services/shared.service.ts
@@ -1,0 +1,43 @@
+import { Injectable } from '@angular/core';
+import { App } from '@capacitor/app';
+import { PopoverController } from '@ionic/angular';
+import { from } from 'rxjs';
+import { switchMap, tap } from 'rxjs/operators';
+import { PopupAlertComponentComponent } from 'src/app/shared/components/popup-alert-component/popup-alert-component.component';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SharedService {
+  constructor(private popoverController: PopoverController) {}
+
+  showAppCloseAlert() {
+    const exitAppPopover = this.popoverController.create({
+      componentProps: {
+        title: 'Exit Fyle App',
+        message: 'Are you sure you want to exit the app?',
+        primaryCta: {
+          text: 'OK',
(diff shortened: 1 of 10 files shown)


src/app/fyle/dashboard/dashboard.page.ts
```diff
@@ -141,6 +158,7 @@ export class DashboardPage implements OnInit {
       relativeTo: this.activatedRoute,
       queryParams,
     });
+    this.registerBackButtonAction();
```
Julias0, line 161: Not super clear what is happening here - might want to add a comment

Julias0 (changes requested):
minor comments

src/app/fyle/dashboard/dashboard.page.ts
yksolanki9: Moved all logic related to `my_dasboard` route in dashboard page and added comments.

src/app/core/services/shared.service.ts
```diff
@@ -0,0 +1,43 @@
...
+import { App } from '@capacitor/app';
+import { PopoverController } from '@ionic/angular';
+import { from } from 'rxjs';
+import { switchMap, tap } from 'rxjs/operators';
+import { PopupAlertComponentComponent } from 'src/app/shared/components/popup-alert-component/popup-alert-component.component';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SharedService {
```
Julias0, line 11: In our application we have a module named Shared, people might confuse it with something related. Can we rename it to BackButtonService or something similar?

Julias0 (approved):
Minor comment. - fix and merge

src/app/core/services/shared.service.ts
yksolanki9: Renamed to BackButtonService

PR #1484, opened 2022-11-10
Code coverage: Org settings service unit tests

Dimple16:
Screenshot 2022-11-18 at 12 52 13 PM


src/app/core/services/org-settings.service.ts
```diff
@@ -16,7 +16,7 @@ export class OrgSettingsService {
   @Cacheable({
     cacheBusterObserver: orgSettingsCacheBuster$,
   })
-  get(): Observable<OrgSettings> {
+  get(): Observable<Partial<OrgSettings>> {
```
Julias0, line 19: Why is this partial? All keys in the response of OrgSettings are not optional right?

src/app/core/services/org-settings.service.spec.ts
```diff
@@ -1,16 +1,412 @@
 import { TestBed } from '@angular/core/testing';
+import { of } from 'rxjs';
+import { EmailEvents, OrgSettings } from '../models/org-settings.model';
+import { ApiService } from './api.service';
 
 import { OrgSettingsService } from './org-settings.service';
 
-xdescribe('OrgSettingsService', () => {
-  let service: OrgSettingsService;
+const apiResponse: Partial<OrgSettings> = {
```
Julias0, line 8: move this to a different file in the same folder

src/app/core/services/org-settings.service.spec.ts
```diff
@@ -1,16 +1,412 @@
...
   });
 
   it('should be created', () => {
-    expect(service).toBeTruthy();
+    expect(orgSettingsService).toBeTruthy();
+  });
+
+  it('should be able to fetch the org settings', (done) => {
+    apiService.get.and.returnValue(of(apiResponse));
+    orgSettingsService.get().subscribe((res) => {
```
Julias0, line 407: If what we get from the api response is the same as what we get from the get method, what do we have the `processIncoming` method?

Julias0 (changes requested):
Added comments

src/app/core/services/org-settings.service.spec.ts
Dimple16: I have handled this in the models and test data responses

src/app/core/services/org-settings.service.spec.ts
```diff
@@ -1,16 +1,131 @@
...
 
   it('should be created', () => {
-    expect(service).toBeTruthy();
+    expect(orgSettingsService).toBeTruthy();
+  });
+
+  it('should be able to get the org settings properly', (done) => {
+    apiService.get.and.returnValue(of(postApiData));
+    orgSettingsService.get().subscribe((res) => {
+      expect(res).toBeTruthy(postApiData);
```
Julias0, line 67: you have used `.toBeTruthy` in a number of places. This doesn't compare by equality but just checks if the response is a truth value or not.

We should add assertions based on what we want to test rather than just for code coverage - same for the rest of the tests

Julias0 (changes requested):
Added important comments

src/app/core/services/org-settings.service.spec.ts
Dimple16: this is done and resolved in all places

Julias0 (approved):

PR #1485, opened 2022-11-11
Code Coverage: Filters-Helper Service

jayfyle:
code-coverage-2

Edit 1: The code coverage is now accurate

Edit 2: Code coverage improved
code-coverage-94

The code coverage is now close to 95%

Edit 3: Code coverage now 100%
code_coverage_100

### DIFF
--- a/src/app/core/models/filters.model.ts
+++ b/src/app/core/models/filters.model.ts
@@ -1,0 +1,9 @@
+import { AdvancesStates } from './advances-states.model';
+import { SortingParam } from './sorting-param.model';
+import { SortingDirection } from './sorting-direction.model';
+
+export type Filters = Partial<{
+  state: AdvancesStates[];
+  sortParam: SortingParam;
+  sortDir: SortingDirection;
+}>;
--- a/src/app/core/models/filters.model.ts
+++ b/src/app/core/models/filters.model.ts
@@ -2,8 +2,8 @@ import { AdvancesStates } from './advances-states.model';
 import { SortingParam } from './sorting-param.model';
 import { SortingDirection } from './sorting-direction.model';
 
-export type Filters = Partial<{
-  state: AdvancesStates[];
-  sortParam: SortingParam;
-  sortDir: SortingDirection;
-}>;
+export interface Filters {
+  state?: AdvancesStates[];
+  sortParam?: SortingParam;
+  sortDir?: SortingDirection;
+}
--- a/src/app/core/services/filters-helper.service.spec.ts
+++ b/src/app/core/services/filters-helper.service.spec.ts
@@ -3,48 +3,184 @@ import { FiltersHelperService } from './filters-helper.service';
 import { Filters } from '../models/filters.model';
 import { AdvancesStates } from '../models/advances-states.model';
 import { FilterPill } from 'src/app/shared/components/fy-filter-pills/filter-pill.interface';
+import { ModalController } from '@ionic/angular';
+import { TitleCasePipe } from '@angular/common';
+import { SortingDirection } from '../models/sorting-direction.model';
+import { SortingParam } from '../models/sorting-param.model';
+import { SelectedFilters } from 'src/app/shared/components/fy-filters/selected-filters.interface';
 
-xdescribe('FiltersHelperService', () => {
+describe('FiltersHelperService', () => {
   let service: FiltersHelperService;
+  let controller: jasmine.SpyObj<ModalController>;
+  let titlePipe: jasmine.SpyObj<TitleCasePipe>;
 
   beforeEach(() => {
-    TestBed.configureTestingModule({});
+    const controllerSpy = jasmine.createSpyObj('ModalController', ['create']);
+    const titleSpy = jasmine.createSpyObj('TitleCasePipe', ['transform']);
+
+    TestBed.configureTestingModule({
+      providers: [
+        {
+          provide: ModalController,
+          useValue: controllerSpy,
+        },
+        {
+          provide: TitleCasePipe,
+          useValue: titleSpy,
+        },
+      ],
+    });
     service = TestBed.inject(FiltersHelperService);
+    controller = TestBed.inject(ModalController) as jasmine.SpyObj<ModalController>;
+    titlePipe = TestBed.inject(TitleCasePipe) as jasmine.SpyObj<TitleCasePipe>;
   });
 
   it('should be created', () => {
     expect(service).toBeTruthy();
   });
 
-  it('should generate filter pills using only state APPROVED defined in filters', () => {
+  /**
+   * Testing Function: generateFilterPills
+   */
+
+  it('should generate pill using only Sorting Params - Approved At New To Old', () => {
     const testFilters: Filters = {
-      state: [AdvancesStates.approved],
+      sortParam: SortingParam.approvalDate,
     };
 
-    const TestPill: FilterPill[] = [
-      {
-        label: 'State',
-        type: 'state',
-        value: 'Approved',
-      },
-    ];
+    const testPill: FilterPill[] = [{ label: 'Sort By', type: 'sort', value: 'approved at - new to old' }];
+
     expect(service.generateFilterPills(testFilters)).toBeTruthy();
-    expect(service.generateFilterPills(testFilters)).toEqual(TestPill);
+    expect(service.generateFilterPills(testFilters)).toEqual(testPill);
   });
 
-  it('should generate filter pills using only state DRAFT defined in filters', () => {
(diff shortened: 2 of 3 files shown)


src/app/core/models/filters.model.ts
```diff
@@ -0,0 +1,9 @@
+import { AdvancesStates } from './advances-states.model';
+import { SortingParam } from './sorting-param.model';
+import { SortingDirection } from './sorting-direction.model';
+
+export type Filters = Partial<{
```
Julias0, line 5: The Partial should added where this is used, not in the definition. 
And when declaring models, use interface instead of type

src/app/core/services/filters-helper.service.spec.ts
```diff
@@ -1,16 +1,291 @@
...
+   */
+
+  it('should generated pill using project name and states', () => {
+    const testFilters: Filters = {
+      state: [AdvancesStates.approved, AdvancesStates.cancelled],
+      sortParam: SortingParam.approvalDate,
+    };
+
+    const result = service.generateFilterPills(testFilters, 'some project');
+    expect(result).toBeTruthy();
```
Julias0, line 56: The check for truthiness is not really required anywhere. Just check if the function behaves as expected and the output of the function is as expected

Julias0: The `some project` argument that is passed is an optional argument. If its not relevant for this test case, lets not add it

src/app/core/services/filters-helper.service.spec.ts
```diff
@@ -1,16 +1,291 @@
...
+  });
+
+  it('should generate pill using only Sorting Params - Approved At New To Old', () => {
```
Julias0, line 59: should generate pill using only Sorting Params - Approved At New To Old
to 
should generate filter pills based on filters selected for sorting on approved at param and no sort direction

Make the sentences descriptive to indicate what we are testing for

Julias0 (changes requested):
Have added a few comments. Make sure these are taken care of in all the test cases

src/app/core/services/filters-helper.service.spec.ts
```diff
@@ -1,16 +1,409 @@
...
+    titlePipe = TestBed.inject(TitleCasePipe) as jasmine.SpyObj<TitleCasePipe>;
   });
 
   it('should be created', () => {
     expect(service).toBeTruthy();
   });
+
+  /**
+   * Testing Function: generateFilterPills
+   */
```
Julias0, line 47: No need to create sections like this - this would be difficult to maintain later on

src/app/core/services/filters-helper.service.spec.ts
```diff
@@ -1,16 +1,409 @@
...
+
+  it('should generated pill using project name and states', () => {
+    const testFilters: Filters = {
+      state: [AdvancesStates.approved, AdvancesStates.cancelled],
+      sortParam: SortingParam.approvalDate,
+    };
+
+    const result = service.generateFilterPills(testFilters, 'some project');
```
Julias0, line 55: why is there to expect statement in this test case? Every test case should have a assertion / expect statement.
If you are just testing for existence of the response that too is fine

src/app/core/services/filters-helper.service.spec.ts
```diff
@@ -1,16 +1,400 @@
...
+        name: 'Sort Direction',
+        value: 1,
+      },
+      {
+        name: 'State',
+        value: ['APPROVED', 'DRAFT'],
+      },
+    ];
+
+    const resFilters: Filters = {
```
Dimple16, line 338: rename this with context, since this is used as expected value, can be `expectedFilters`?

Dimple16 (approved):

src/app/core/services/filters-helper.service.spec.ts
```diff
@@ -1,16 +1,400 @@
...
     service = TestBed.inject(FiltersHelperService);
+    controller = TestBed.inject(ModalController) as jasmine.SpyObj<ModalController>;
+    titlePipe = TestBed.inject(TitleCasePipe) as jasmine.SpyObj<TitleCasePipe>;
   });
 
   it('should be created', () => {
     expect(service).toBeTruthy();
   });
+
+  it('should generated pill using project name', () => {
```
Julias0, line 45: This description is not correct right? You are generating a pill based on the approval date.

src/app/core/services/filters-helper.service.spec.ts
```diff
@@ -1,16 +1,400 @@
...
+import { SortingValue } from '../models/sorting-value.model';
+import { ModalController } from '@ionic/angular';
+import { TitleCasePipe } from '@angular/common';
+import { SortingDirection } from '../models/sorting-direction.model';
+import { SortingParam } from '../models/sorting-param.model';
+import { SelectedFilters } from 'src/app/shared/components/fy-filters/selected-filters.interface';
 
-xdescribe('FiltersHelperService', () => {
+describe('FiltersHelperService', () => {
   let service: FiltersHelperService;
```
Julias0, line 16: rename to filterHelperService

src/app/core/services/filters-helper.service.spec.ts
```diff
@@ -1,16 +1,400 @@
...
+  let controller: jasmine.SpyObj<ModalController>;
```
Julias0, line 17: rename to modalController

src/app/core/services/filters-helper.service.spec.ts
```diff
@@ -1,16 +1,400 @@
...
+    };
+
+    const testSelectedFilter: SelectedFilters<any>[] = [
+      {
+        name: 'Sort By',
+        value: 'appDateNewToOld',
+      },
+      {
+        name: 'Sort Direction',
+        value: 1,
```
Julias0, line 160: convert SortingDirection.descending to a number to get the value here. Otherwise this case will become flaky

src/app/core/services/filters-helper.service.spec.ts
```diff
@@ -1,16 +1,400 @@
...
+    };
+
+    const testSelectedFilter: SelectedFilters<any>[] = [
+      {
+        name: 'Sort By',
+        value: 'projectZToA',
+      },
+      {
+        name: 'Sort Direction',
+        value: 1,
```
Julias0, line 212: Same as above

src/app/core/services/filters-helper.service.spec.ts
```diff
@@ -1,16 +1,400 @@
...
+          {
+            label: 'Approved At - Old to New',
+            value: SortingValue.approvalDateDesc,
+          },
+        ],
+      },
+    ];
+
+    controller.create.and.returnValue(
+      new Promise((resolve) => {
```
Julias0, line 385:
```typescript

const filterPopoverResponse = new Promise((resInt) => {
            resInt({
              data: selectedFilters,
            });
          });
const modalControllerResponse =  new Promise((resolve) => {
const filterPopoverSpy = jasmine.createSpyObj('filterPopover', ['onWillDismiss', 'present']) as any;
        filterPopoverSpy.onWillDismiss.and.returnValue(
          filterPopoverResponse
        );
        resolve(filterPopoverSpy);
      })

```

Julias0: `controller.create.and.returnValue(modalControllerResponse);

Julias0 (changes requested):
Added more comments

src/app/core/services/filters-helper.service.spec.ts
```diff
@@ -1,16 +1,406 @@
...
+    const testFilters: Filters = {
+      state: [AdvancesStates.approved, AdvancesStates.draft],
+      sortParam: SortingParam.approvalDate,
+    };
+
+    const filterResponse = [
+      {
+        label: 'State',
+        type: 'state',
+        value: ', ',
```
Julias0, line 55: Wait - this seems to be a bug?

src/app/core/services/filters-helper.service.spec.ts
```diff
@@ -1,16 +1,406 @@
...
+      },
+      {
+        label: 'Sort By',
+        type: 'sort',
+        value: 'approved at - new to old',
+      },
+    ];
+
+    expect(filterHelperService.generateFilterPills(testFilters, 'some project')).toEqual(filterResponse);
```
Julias0, line 64: Again, we havent passed a filter for project right? If yes, we are not truly testing for the `'some project'` value which is passed in

Julias0 (changes requested):
Added important comments

Julias0 (approved):

Dimple16 (approved):

PR #1498, opened 2022-11-17
Code Coverage: Status Service

jayfyle:
## Status Service Code Coverage Report:

code-coverage-service

Edit 1: Unit test for status map and find all comments function added. Some functions not being used anywhere were removed.

Edit 2: Unit test for post and find latest comment function added, covering all functions in the service

status-service-100-coverage


src/app/core/test-data/status.service.spec.data.ts
```diff
@@ -0,0 +1,1274 @@
...
+    st_transaction_id: null,
+    st_report_id: 'rp39MrwyOm9n',
+    st_advance_request_id: null,
+    st_trip_request_id: null,
+    us_full_name: null,
+    us_email: null,
+    isBotComment: true,
+    isSelfComment: false,
+    isOthersComment: true,
+    st: {
```
Julias0, line 111: wait, I dont think we send the category and icon from the API response right? This is not a right sample of the API response

src/app/core/test-data/status.service.spec.data.ts
```diff
@@ -0,0 +1,1162 @@
...
+    st_trip_request_id: null,
+    us_full_name: null,
+    us_email: null,
+    isBotComment: true,
+    isSelfComment: false,
+    isOthersComment: true,
+  },
+];
+
+export const updateResponseWithSt: ExtendedStatus[] = [
```
Dimple16, line 573: `updateReponseWithFlattenedEStatus`

src/app/core/test-data/status.service.spec.data.ts
```diff
@@ -0,0 +1,1162 @@
+import { ExtendedStatus } from '../models/extended_status.model';
+
+export const apiResponse: ExtendedStatus[] = [
```
Dimple16, line 3: rename this to `getEStatusApiResponse`

src/app/core/test-data/status.service.spec.data.ts
```diff
@@ -0,0 +1,1162 @@
...
+    st_transaction_id: null,
+    st_report_id: 'rp39MrwyOm9n',
+    st_advance_request_id: null,
+    st_trip_request_id: null,
+    us_full_name: null,
+    us_email: null,
+    isBotComment: true,
+    isSelfComment: false,
+    isOthersComment: true,
+    st: {
```
Dimple16, line 590: we can remove the `st:` from here since this is not passed by the client, we just use it in the client for our grouping, this isn't stored in the backend

Dimple16 (changes requested):
add the updated code coverage report screenshot, the current one doesn't have 100% coverage, else, you can add the test cases and raise PR again

Julias0 (changes requested):
1 minor comment

Dimple16 (approved):

PR #1502, opened 2022-11-18
Remove LD flag for payment_mode_configurations and automated report submissions

arjunaj5:
https://app.clickup.com/t/30k7g3x
https://app.clickup.com/t/30k7ewy

### DIFF
--- a/src/app/core/services/launch-darkly.service.ts
+++ b/src/app/core/services/launch-darkly.service.ts
@@ -72,10 +72,6 @@ export class LaunchDarklyService {
     return this.getVariation('hide_paid_by_company', false);
   }
 
-  checkIfAutomateReportSubmissionIsEnabled() {
-    return this.getVariation('automate_report_submission_enabled', false);
-  }
-
   // Checks if the passed in user is the same as the user which is initialized to LaunchDarkly (if any)
   private isTheSameUser(newUser: LDClient.LDUser): boolean {
     const previousUser = this.ldClient?.getUser();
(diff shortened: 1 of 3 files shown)


gokulk16 (approved):

src/app/core/services/payment-modes.service.ts
```diff
@@ -26,14 +26,11 @@ export class PaymentModesService {
 
   checkIfPaymentModeConfigurationsIsEnabled() {
     return forkJoin({
```
yksolanki9, line 28: forkJoin is not needed here

src/app/core/services/launch-darkly.service.ts
```diff
@@ -68,18 +68,10 @@ export class LaunchDarklyService {
     return this.getVariation('keyboard_plugin_enabled', true);
   }
 
-  checkIfPaymentModeConfigurationsIsEnabled(): Observable<boolean> {
```
yksolanki9, line 71: This method is being used in many places, you need to remove it from everywhere

yksolanki9 (changes requested):

src/app/core/services/launch-darkly.service.ts
arjunaj5: There is one more method in `PaymentModesService` with the same name, It is being used in other places. 
This method was used only by `PaymentModesService.checkIfPaymentModeConfigurationsIsEnabled()` which is removed in this PR

src/app/core/services/payment-modes.service.ts
arjunaj5: fixed it

yksolanki9 (approved):
Please test this with and without the PMS setting enabled before merging

PR #1503, opened 2022-11-18
Added tests for tasks service

Julias0:
Screenshot 2022-11-22 at 5 41 25 PM


src/app/core/services/tasks.service.ts
```diff
@@ -53,9 +51,13 @@ export class TasksService {
...
     private currencyService: CurrencyService
   ) {
+    this.refreshOnTaskClear();
+  }
+
+  refreshOnTaskClear(): void {
     this.userEventService.onTaskCacheClear(() => {
       this.reportService.getReportAutoSubmissionDetails().subscribe((autoSubmissionReportDetails) => {
-        const isReportAutoSubmissionScheduled = !!autoSubmissionReportDetails?.data?.next_at;
+        const isReportAutoSubmissionScheduled = false;
```
Julias0, line 60: remove

src/app/core/services/tasks.service.ts
```diff
@@ -53,6 +51,10 @@ export class TasksService {
     private advancesRequestService: AdvanceRequestService,
     private currencyService: CurrencyService
   ) {
+    this.refreshOnTaskClear();
+  }
+
+  refreshOnTaskClear(): void {
```
Julias0, line 57: separating this method made it much easier to unit test this. This is completely ok to do - if you find a piece of code difficult to reach, extract it out - this might make it much much easier to test.

src/app/core/services/tasks.service.ts
```diff
@@ -218,17 +220,17 @@ export class TasksService {
 
   getExpensePill(filters: TaskFilters): FilterPill {
     const expensePills = [];
-    const draftExpensesContent = filters.draftExpenses ? 'Incomplete' : '';
-    if (draftExpensesContent) {
-      expensePills.push(draftExpensesContent);
+
+    if (filters.draftExpenses) {
```
Julias0, line 224: Once your code is covered by unit tests - reactors like this become very simple to do. There was an edge case here which would have never happened, so I first covered it with unit tests and then refactored it.
This gives confidence that the external behaviour of the method hasn't changed

src/app/core/services/tasks.service.ts
```diff
@@ -357,7 +359,7 @@ export class TasksService {
     );
   }
 
-  private getFilteredTaskList(filters: TaskFilters, tasksDict: TaskDict) {
+  getFilteredTaskList(filters: TaskFilters, tasksDict: TaskDictionary): DashboardTask[] {
```
Julias0, line 362: If a method is marked by private, it may become very difficult to unit test it. You might find your unit tests becoming very complicated or confusing because an internal method has a lot of branching etc.

It is completely fine to remove private to make your code easier to test.

src/app/core/services/tasks.service.ts
```diff
@@ -400,15 +402,15 @@ export class TasksService {
     return tasks;
   }
 
-  private getSentBackReports() {
+  getSentBackReports() {
```
Julias0, line 405: havent added data types everywhere - will do separately, this PR is too big as is

src/app/core/services/tasks.service.spec.ts
```diff
@@ -1,16 +1,1048 @@
...
+    expect(tasksService).toBeTruthy();
+  });
+
+  it('should be able to fetch tasks related to sent back advances', (done) => {
+    currencyService.getHomeCurrency.and.returnValue(of(homeCurrency));
+    advanceRequestService.getMyAdvanceRequestStats.and.returnValue(of(sentBackAdvancesResponse));
+    humanizeCurrencyPipe.transform
+      .withArgs(sentBackAdvancesResponse[0].aggregates[1].function_value, homeCurrency, true)
+      .and.returnValue('123.37M');
+    humanizeCurrencyPipe.transform
```
Julias0, line 216: Why not use the actual pipe? Because dependencies will convert this from a unit test to an integration test. What if there is a bug inside `humanizeCurrencyPipe` which causes these tests to pass or fail?

Unit tests should function in as much isolation as possible as a best practice

src/app/core/services/tasks.service.spec.ts
```diff
@@ -1,16 +1,1048 @@
...
+      .withArgs(sentBackAdvancesResponse[0].aggregates[1].function_value, homeCurrency)
+      .and.returnValue('₹123.37M');
+
+    tasksService.getSentBackAdvanceTasks().subscribe((sentBackAdvancesData) => {
+      expect(sentBackAdvancesData).toEqual([sentBackAdvanceTaskSample]);
+      done();
+    });
+  });
+
+  function setupUnsibmittedReportsResponse() {
```
Julias0, line 226: why is there a function like this here?
Your unit testing code is still code - if you find it getting unclean due to a lot of complicate statements, you can always refactor it and break it down using functions. You don't have to write everything inside the `it` block, you can have reusable code. 

Just make sure you do the assertions inside the `it` blocks, everything else can be extracted and moved

src/app/core/models/task-event.enum.ts
```diff
@@ -1,10 +1,10 @@
 export enum TASKEVENT {
-  expensesCreateNewReport,
-  expensesAddToReport,
-  openSentBackReport,
-  reviewExpenses,
-  openDraftReports,
-  openTeamReport,
-  openPotentialDuplicates,
-  openSentBackAdvance,
+  expensesCreateNewReport = 0,
```
Julias0, line 2: Not assigning numerical values can cause the number associated with each value to change if someone adds a new enum to the top or middle of the declared values

src/app/shared/components/capture-receipt/camera-preview/camera-preview.component.spec.ts
```diff
@@ -7,16 +7,18 @@ xdescribe('CameraPreviewComponent', () => {
   let component: CameraPreviewComponent;
   let fixture: ComponentFixture<CameraPreviewComponent>;
 
-  beforeEach(waitForAsync(() => {
-    TestBed.configureTestingModule({
-      declarations: [CameraPreviewComponent],
-      imports: [IonicModule.forRoot()],
-    }).compileComponents();
+  beforeEach(
```
Julias0, line 10: ignore - indentation

Dimple16 (approved):

PR #1511, opened 2022-11-23
#mission_nps_mobile_40: Fix view report page not updating after adding expenses to report

yksolanki9:
Fixes:
- When adding expense to report, the expenses inside report were not updated
- After adding expense to report, the unreported expenses modal is not updated which results in 400 errors
- Removed unnecessary logic and did some refactoring

UI Preview:
https://user-images.githubusercontent.com/39493102/203567708-372e1ad6-f3ff-4849-8300-936c6092eeab.mov


Dimple16 (approved):

src/app/fyle/my-view-report/my-view-report.page.ts
```diff
@@ -180,20 +162,22 @@ export class MyViewReportPage {
...
     this.setupNetworkWatcher();
-    this.isExpensesLoading = true;
+    this.reportId = this.activatedRoute.snapshot.params.id;
     this.navigateBack = !!this.activatedRoute.snapshot.params.navigateBack;
 
-    this.erpt$ = from(this.loaderService.showLoader()).pipe(
-      switchMap(() => this.reportService.getReport(this.activatedRoute.snapshot.params.id)),
-      finalize(() => from(this.loaderService.hideLoader()))
+    this.erpt$ = this.loadReportDetails$.pipe(
+      switchMap(() => this.loaderService.showLoader()),
```
Julias0, line 169: why use a switchMap here?

src/app/fyle/my-view-report/my-view-report.page.ts
```diff
@@ -180,20 +162,22 @@ export class MyViewReportPage {
...
+      switchMap(() => this.reportService.getReport(this.reportId)),
+      tap(() => from(this.loaderService.hideLoader())),
```
Julias0, line 171: This should be done in finalize. Otherwise this will not get executed if the API call fails

src/app/fyle/my-view-report/my-view-report.page.ts
```diff
@@ -234,35 +218,26 @@ export class MyViewReportPage {
...
-    this.erpt$.subscribe((erpt) => {
+    this.erpt$.pipe(take(1)).subscribe((erpt) => {
       this.reportCurrencySymbol = getCurrencySymbol(erpt?.rp_currency, 'wide');
-      this.reportName = erpt?.rp_purpose;
-      this.reportId = erpt?.rp_id;
 
       //For sent back reports, show the comments section instead of expenses when opening the report
       if (erpt.rp_state === 'APPROVER_INQUIRY') {
-        this.ionSegment.value = 'comments';
+        setTimeout(() => (this.ionSegment.value = 'comments'), 0);
```
Julias0, line 226: Why is this change required?

src/app/fyle/my-view-report/my-view-report.page.ts
Julias0, line 171: This should be inside finalize.

src/app/fyle/my-view-report/my-view-report.page.ts
```diff
@@ -274,10 +249,10 @@ export class MyViewReportPage {
         })
       ),
       shareReplay(1),
-      finalize(() => (this.isExpensesLoading = false))
+      tap(() => (this.isExpensesLoading = false))
```
Julias0, line 252: This should be inside finalize.

Julias0 (changes requested):
minor comments

src/app/fyle/my-view-report/my-view-report.page.ts
yksolanki9: If I remove this, there's an error - `Reading value of undefined`. There's only one ngIf in the parent element where we use the async pipe for `erpt$`.
So, this is probably bc the value was emitted but the child elements have not been added to the DOM yet. Adding setTimeout fixes it.

yksolanki9: Let me know if you have something else in mind

src/app/fyle/my-view-report/my-view-report.page.ts
yksolanki9: `finalize` is called only when the source observable completes, but here, `loadReportDetails$` is a BehaviourSubject which'll emit a new value everytime there's an update

src/app/fyle/my-view-report/my-view-report.page.ts
yksolanki9: `finalize` is called only when the source observable completes, but here, `loadReportTxns$` is a BehaviourSubject which'll emit a new value everytime there's an update

src/app/fyle/my-view-report/my-view-report.page.ts
yksolanki9: Fixed

src/app/fyle/my-view-report/my-view-report.page.ts
yksolanki9: Fixed as discussed

src/app/fyle/my-view-report/my-view-report.page.ts
yksolanki9: Fixed as discussed

src/app/fyle/my-view-report/my-view-report.page.ts
Julias0: cant we just set value as a input type property and set it via a variable? Why do we need to do this viewChild stuff?

<ion-segment [value]="selectedSegment" >

Julias0 (changes requested):
1 comment

Julias0 (approved):

PR #1513, opened 2022-11-29
Code coverage: Accounts Service

Dimple16:


src/app/core/services/accounts.service.ts
```diff
@@ -46,47 +46,18 @@ export class AccountsService {
...
     const { etxn, orgSettings, expenseType, isPaymentModeConfigurationsEnabled, isPaidByCompanyHidden } = config;
-    const isAdvanceEnabled = orgSettings?.advances?.enabled || orgSettings?.advance_requests?.enabled;
-    const isMultipleAdvanceEnabled = orgSettings?.advance_account_settings?.multiple_accounts;
+    const isAdvanceEnabled = orgSettings.advances.enabled || orgSettings.advance_requests.enabled;
+    const isMultipleAdvanceEnabled = orgSettings.advance_account_settings.multiple_accounts;
     const isMileageOrPerDiemExpense = [ExpenseType.MILEAGE, ExpenseType.PER_DIEM].includes(expenseType);
 
-    let userAccounts = this.filterAccountsWithSufficientBalance(accounts, isAdvanceEnabled);
-
-    if (!isPaymentModeConfigurationsEnabled) {
```
Dimple16, line 62: since payment mode configurations are live for all orgs on prod now, this block is removed, not useful anymore

src/app/core/services/accounts.service.ts
```diff
@@ -131,116 +102,39 @@ export class AccountsService {
...
       (account) =>
         // Personal Account and CCC account are considered to always have sufficient funds
         (isAdvanceEnabled && account.acc.tentative_balance_amount > 0) ||
-        [AccountType.PERSONAL, AccountType.CCC].indexOf(account.acc.type) > -1 ||
-        accountId === account.acc.id
+        [AccountType.PERSONAL, AccountType.CCC].indexOf(account.acc.type) > -1
     );
   }
 
-  private constructPaymentModes(
```
Dimple16, line 194: removed unused method which was related to payment mode configurations disabled

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -1456,9 +1456,6 @@ export class AddEditExpensePage implements OnInit {
           return CCCAccount.value;
         }
 
-        if (!isPaymentModeConfigurationsEnabled) {
```
Dimple16, line 1459: unused, live for everyone now

src/app/core/models/report-unflattened.model.ts
```diff
@@ -4,40 +4,40 @@ import { User } from './user.model';
 export interface UnflattenedReport {
   ou: OrgUser;
   rp: {
```
Julias0, line 6: You can use the Partial Generic class to avoid adding optional to all the properties

src/app/core/models/unflattened-transaction.model.ts
```diff
@@ -1,13 +1,25 @@
...
+import { UnflattenedReport } from './report-unflattened.model';
 import { User } from './user.model';
 import { Transaction } from './v1/transaction.model';
 
 export interface UnflattenedTransaction {
-  ou: OrgUser;
-  tx: Transaction;
-  us: User;
-  source: {
+  ou?: OrgUser;
```
Julias0, line 7: Why are all these properties marked optional? A bunch of these will always be there no

src/app/core/services/accounts.service.spec.ts
```diff
@@ -2,117 +2,93 @@ import { TestBed } from '@angular/core/testing';
...
+
+const multiplePaymentModesWithoutCCCAcc = multiplePaymentModesWithoutCCCAccData;
+
+const etxnObjWithSource = etxnObjWithSourceData;
+
+const paymentModesRes = paymentModesResData;
+
+const paymentModesAccounts = paymentModesAccountsData;
+
+const multiplePaymentModesIncPersonalAcc = multiplePaymentModesIncPersonalAccData;
```
Julias0, line 90: Why not use the imported variables directly? Why is this assignment to an intermediate variable required?

src/app/core/models/v1/transaction.model.ts
```diff
@@ -1,74 +1,95 @@
...
+  project_name?: string;
+  project_code?: string;
+  cost_center_name?: string;
+  cost_center_code?: string;
+  transcription_state?: string;
+  verification_state?: string;
+  org_category?: string;
+  sub_category?: string;
+  org_category_code?: string;
+  transcribed_data?: {};
```
Julias0, line 88: transcribed_data has a data model na - can we use that

src/app/core/models/v1/transaction.model.ts
```diff
@@ -1,74 +1,95 @@
 export interface Transaction {
```
Julias0, line 1: We should use the Partial generic class in places where the properties can be optionally present.
Example - 
When we do a get call and retrieve a transaction a lot of these fields will be mandatorily present. But when we do a post call during creation the properties become optional. Both cases doesn't necessarily share the same model so we should be using the one which is valid on a case to case basis

src/app/core/services/accounts.service.spec.ts
```diff
@@ -154,14 +80,188 @@ describe('AccountsService', () => {
...
     apiService.get.and.returnValue(of(accountsCallResponse1));
-    dataTransformService.unflatten.withArgs(account1).and.returnValue(unflattenedAccount1);
-    dataTransformService.unflatten.withArgs(account2).and.returnValue(unflattenedAccount2);
+    dataTransformService.unflatten.withArgs(account1Data).and.returnValue(unflattenedAccount1Data);
+    dataTransformService.unflatten.withArgs(account2Data).and.returnValue(unflattenedAccount2Data);
 
     accountsService.getEMyAccounts().subscribe((res) => {
-      expect(res[0]).toEqual(unflattenedAccount1);
-      expect(res[1]).toEqual(unflattenedAccount2);
+      expect(res[0]).toEqual(unflattenedAccount1Data);
```
Julias0, line 87: unflattenedAccount1Data has the date in the format of Date object. This is not handled inside the actual method definition

src/app/core/services/accounts.service.spec.ts
```diff
@@ -154,14 +80,188 @@ describe('AccountsService', () => {
...
     });
   });
+
+  it('should be able to check if etxn has same payment mode', () => {
+    expect(accountsService.checkIfEtxnHasSamePaymentMode(unflattenedTransactionCCC, paymentModeDataCCC)).toEqual(true);
+  });
+
+  it('should be able to check if etxn has same personal account payment mode', () => {
+    expect(
+      accountsService.checkIfEtxnHasSamePaymentMode(unflattenedTransactionPersonal, paymentModeDataPersonal)
```
Julias0, line 100:
```
acc: {
    id: 'accZ1IWjhjLyu4',
    isReimbursable: false,
  },
  source: {
    account_type: AccountType.PERSONAL,
    account_id: 'accZ1IWjhjLv4',
  }
  ```
  
  ```
acc: {
    id: 'accZ1IWjhjLyu4',
    name: 'Personal Account',
    type: AccountType.PERSONAL,
    currency: 'INR',
    target_balance_amount: 0,
    current_balance_amount: 0,
    tentative_balance_amount: 0,
    category: null,
    displayName: 'Personal',
    isReimbursable: false,
  }
  ```
We should also check for the true case for this test case, otherwise if this case breaks it wont come up

File: src/app/core/services/accounts.service.spec.ts [hunk 11]
```diff
@@ -154,14 +80,188 @@ describe('AccountsService', () => {
...
+    ).toEqual(false);
+  });
+
+  it('should be able to get etxn selected payment mode with source account id', () => {
+    expect(accountsService.getEtxnSelectedPaymentMode(unflattenedTransactionCCC, paymentModesData)).toEqual(
+      paymentModeDataCCC
+    );
+  });
+
+  it('should be able to get etxn selected payment mode null without source account id', () => {
```
Julias0, line 110: should be able to get selected payment mode `as` null when extn  is without source account id

src/app/core/services/accounts.service.spec.ts
Dimple16: added the `fixDates` to accounts service

Dimple16: this was old data, I've updated it now, and the model has the type as `Date`, it was a glitch in the spec.data - fixed now

src/app/core/services/accounts.service.spec.ts
Dimple16: added another expect for true case

PR #1519, opened 2022-11-30
#mission_nps_mobile_40: Use app font size irrespective of device native font size

yksolanki9:
- I'm using capacitor's official `text-zoom` plugin here - https://capacitorjs.com/docs/apis/text-zoom
- It has been around for a while and already has support for Capacitor 4
- Tested on both android and IOS 🟢 
- Will get a green from Sumanth and Bhavika before merging and we'll do a staged release as discussed

### DIFF
--- a/android/app/capacitor.build.gradle
+++ b/android/app/capacitor.build.gradle
@@ -22,12 +22,15 @@ dependencies {
     implementation project(':capacitor-splash-screen')
     implementation project(':capacitor-status-bar')
     implementation project(':capacitor-storage')
+    implementation project(':capacitor-text-zoom')
+    implementation project(':capacitor-native-settings')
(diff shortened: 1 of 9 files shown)


android/app/capacitor.build.gradle
```diff
@@ -22,12 +22,15 @@ dependencies {
     implementation project(':capacitor-splash-screen')
     implementation project(':capacitor-status-bar')
     implementation project(':capacitor-storage')
+    implementation project(':capacitor-text-zoom')
+    implementation project(':capacitor-native-settings')
```
yksolanki9, line 26: I've added only one plugin - `capacitor-text-zoom`
`capacitor-native-settings` - this must have been removed accidentally. We use it for opening native settings for camera/gallery permission

Julias0 (approved):

PR #1524, opened 2022-12-02
#mission_nps_mobile_40: Support HEIC uploads from file explorer

yksolanki9:


src/app/core/services/file.service.ts
```diff
@@ -124,11 +125,44 @@ export class FileService {
...
+  readFile(file: Blob): Promise<string | ArrayBuffer> {
     return new Promise((resolve, reject) => {
       const fileReader = new FileReader();
-      fileReader.onload = () => resolve(fileReader.result);
+      fileReader.onload = async () => {
+        if (file.type === 'image/heic') {
+          const result = await heic2any({
+            blob: this.getBlobFromDataUrl(fileReader.result as string),
+            toType: 'image/jpeg',
+            quality: 50,
```
yksolanki9, line 157: Setting this to 50 since iphones usually have good camera. Verified on my device DE works fine even with 50% compression.

Julias0 (approved):

PR #1568, opened 2022-12-21
Code Coverage : Per-Diem Service

jayfyle:
# Code Coverage For Per Diem Service
## Edit 1 : Code coverage at 100 across the board except for branch coverage

per-diem-service_code_coverage

## Edit 2: Branch coverage improved
image

## Edit 3: Branch Coverage almost 100
Per_diem_service_almost_100

Test_service_per_diem_better_coverage


Dimple16 (changes requested):

src/app/core/services/per-diem.service.spec.ts
```diff
@@ -1,16 +1,104 @@
...
+
+    const result = perDiemService.getRates();
+    result.subscribe((res) => {
+      expect(res).toEqual(fixDate(expectedPerDiemRates));
+      done();
+    });
+  });
+
+  it('should get all allowed per diems', (done) => {
+    orgUserSettingsService.get.and.returnValue(of(apiOrgUserSettings));
```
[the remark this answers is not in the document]
jayfyle: I don't think as the function returns the allowed per diems using the allowed per diems inside the org user settings

src/app/core/services/per-diem.service.spec.ts
```diff
@@ -1,16 +1,104 @@
...
+    spenderPlatformApiService.get.and.returnValue(of(apiPerDiemRates));
+
+    const result = perDiemService.getRates();
+    result.subscribe((res) => {
+      expect(res).toEqual(fixDate(expectedPerDiemRates));
+      done();
+    });
+  });
+
+  it('should get all allowed per diems', (done) => {
```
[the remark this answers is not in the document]
jayfyle: Yes, we are checking for both the cases

Dimple16 (approved):

src/app/core/services/per-diem.service.spec.ts
```diff
@@ -1,16 +1,114 @@
...
+      done();
+    });
+  });
+
+  it('should get per diem rates', (done) => {
+    spenderPlatformApiService.get.and.returnValue(of(apiCountResponse));
+    spenderPlatformApiService.get.and.returnValue(of(apiPerDiemRates));
+
+    const result = perDiemService.getRates();
+    result.subscribe((res) => {
```
Julias0, line 76: we should also add assertions to check if 
`this.spenderPlatformApiService
      .get` is being called the proper number of times and with proper assertions

src/app/core/services/per-diem.service.spec.ts
```diff
@@ -1,16 +1,114 @@
...
+
+    const result = perDiemService.getRate(actualId);
+    result.subscribe((res) => {
+      expect(res).toEqual(expectPerDiemByID);
+      done();
+    });
+  });
+
+  it('should get per diem rates', (done) => {
+    spenderPlatformApiService.get.and.returnValue(of(apiCountResponse));
```
Julias0, line 72: The count response is `count: 57` and the `spenderPlatformApiService.get` is mocked to return 2 counts of data. The data for this test case is incorrect na

src/app/core/services/per-diem.service.spec.ts
```diff
@@ -1,16 +1,114 @@
...
+    spenderPlatformApiService.get.and.returnValue(of(apiPerDiemRates));
```
Julias0, line 73: You have mocked the same thing twice - the second one is overriding the first one. This doesn't test for a case where you have multiple pages of data.

Add a apiCount of 4 and the page side of 2, so that we can check if pulling all of the data is working properly. Also we need to check if `this.spenderPlatformApiService
      .get` is being called with the right arguments otherwise what are we even testing in this test case

src/app/core/test-data/per-diem.service.spec.data.ts
```diff
@@ -0,0 +1,316 @@
...
+    currency: 'INR',
+    id: 502,
+    name: 'Fyle-Test2',
+    org_id: 'orNVthTo2Zyo',
+    rate: 65,
+    updated_at: '2020-11-20T13:36:16.154Z',
+  },
+];
+
+export const allPerDiemRatesParam = [
```
Julias0, line 106: why is this named param? This is response data right?

src/app/core/services/per-diem.service.spec.ts
```diff
@@ -1,16 +1,114 @@
...
+    const result = perDiemService.getAllowedPerDiems(null);
+
+    result.subscribe((res) => {
+      expect(res).toEqual([]);
+      done();
+    });
+  });
+
+  it('should return empty list if there is no settings returned', (done) => {
+    orgUserSettingsService.get.and.returnValue(of(null));
```
Julias0, line 105: I dont think there will be a case where org_user_settings will even be null. The type safety is probably wrongly added for this

Julias0 (changes requested):
Important comments

src/app/core/services/per-diem.service.spec.ts
```diff
@@ -1,16 +1,126 @@
...
+
+    const result = perDiemService.getRate(actualId);
+    result.subscribe((res) => {
+      expect(res).toEqual(expectPerDiemByID);
+      done();
+    });
+  });
+
+  it('should get per diem rates', (done) => {
+    spenderPlatformApiService.get.and.returnValue(of(apiPerDiem));
```
Julias0, line 71: dont we need to mock the responses by call? Otherwise this is returning the same value in each call. 
Example - 
apiPerDiem in the count call gives 3 in count.
But in the  expectedPerDiemRates you are having 6 values which would never be the case

src/app/core/test-data/per-diem.service.spec.data.ts
Dimple16: this is used as param for the method `getAllowedPerDiemRates()`, 
we pass the list of per diem rates and filter the allowed ones

Dimple16 (approved):

Julias0 (approved):

PR #1575, opened 2023-01-02
fixes for the sentry issues

yashfalke77:
Clickup Task: https://app.clickup.com/t/85zrh6cgx
Sentry issues thats been fixed:
1) https://sentry.io/organizations/fyle-technologies-private-limi/issues/3840499445/?project=5622998&referrer=release-issue-stream
2) https://sentry.io/organizations/fyle-technologies-private-limi/issues/3836338631/?project=5622998&referrer=release-issue-stream

### DIFF
--- a/src/app/auth/switch-org/switch-org.page.ts
+++ b/src/app/auth/switch-org/switch-org.page.ts
@@ -179,7 +179,7 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
      * After showing the alert, the user will be redirected to the sign-in page since there is no other org they are a part of.
      * If the user has more than 1 org, the user will stay on the switch org page to choose another org.
      */
-    if (orgs?.length > 1) {
+    if (orgs?.length === 1) {
       this.signOut();
     }
   }
--- a/src/app/auth/switch-org/switch-org.page.ts
+++ b/src/app/auth/switch-org/switch-org.page.ts
@@ -205,12 +205,14 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
 
   async showEmailNotVerifiedAlert() {
     const eou$ = from(this.authService.getEou());
-    forkJoin([eou$, this.orgs$]).subscribe(async (result) => {
-      const orgName = result[0].ou.org_name;
-      const orgId = result[0].ou.org_id;
-      const email = result[0].us.email;
-      this.orgs = result[1];
-
+    forkJoin({
+      eou: eou$,
+      orgs: this.orgs$,
+    }).subscribe(async ({ eou, orgs }) => {
+      const orgName = eou.ou.org_name;
+      const orgId = eou.ou.org_id;
+      const email = eou.us.email;
+      this.orgs = orgs;
       const popover = await this.popoverController.create({
         componentProps: {
           title: 'Invite Not Accepted',
--- a/src/app/auth/switch-org/switch-org.page.ts
+++ b/src/app/auth/switch-org/switch-org.page.ts
@@ -179,7 +179,7 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
      * After showing the alert, the user will be redirected to the sign-in page since there is no other org they are a part of.
      * If the user has more than 1 org, the user will stay on the switch org page to choose another org.
      */
-    if (this.orgs.length === 1) {
+    if (this.orgs?.length === 1) {
       this.signOut();
     }
   }
--- a/src/app/auth/switch-org/switch-org.page.ts
+++ b/src/app/auth/switch-org/switch-org.page.ts
@@ -173,18 +173,18 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
     this.trackingService.showToastMessage({ ToastContent: toastMessageData.message });
   }
 
-  logoutIfSingleOrg() {
+  logoutIfSingleOrg(orgs: Org[]) {
     /*
      * Case: When a user is added to an SSO org but hasn't verified their account through the link.
      * After showing the alert, the user will be redirected to the sign-in page since there is no other org they are a part of.
      * If the user has more than 1 org, the user will stay on the switch org page to choose another org.
      */
-    if (this.orgs?.length === 1) {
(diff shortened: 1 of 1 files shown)


src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -179,12 +179,12 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
      * After showing the alert, the user will be redirected to the sign-in page since there is no other org they are a part of.
      * If the user has more than 1 org, the user will stay on the switch org page to choose another org.
      */
-    if (this.orgs.length === 1) {
+    if (this.orgs?.length === 1) {
```
yksolanki9, line 182: Why do we still need this?

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -204,10 +204,12 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
   }
 
   async showEmailNotVerifiedAlert() {
-    this.authService.getEou().then(async (eou) => {
-      const orgName = eou.ou.org_name;
-      const orgId = eou.ou.org_id;
-      const email = eou.us.email;
+    const eou$ = from(this.authService.getEou());
+    forkJoin([eou$, this.orgs$]).subscribe(async (result) => {
```
yksolanki9, line 208: Use object instead of array

yksolanki9 (changes requested):

yksolanki9 (commented):
https://sentry.io/organizations/fyle-technologies-private-limi/issues/3840499445/?project=5622998&referrer=release-issue-stream

How is this issue fixed?

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -224,12 +226,13 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
...
         component: PopupAlertComponentComponent,
         cssClass: 'pop-up-in-center',
+        backdropDismiss: false,
       });
       await popover.present();
 
       const { data } = await popover.onWillDismiss();
 
-      this.handleDismissPopup(data.action, email, orgId);
+      this.handleDismissPopup(data?.action, email, orgId);
```
Dimple16, line 235: shouldn't the method `handleDismissPopup` be called if `data` is present? 
or, are you handling the null case separately?

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -204,10 +204,12 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
...
-    this.authService.getEou().then(async (eou) => {
-      const orgName = eou.ou.org_name;
-      const orgId = eou.ou.org_id;
-      const email = eou.us.email;
+    const eou$ = from(this.authService.getEou());
+    forkJoin([eou$, this.orgs$]).subscribe(async (result) => {
+      const orgName = result[0].ou.org_name;
+      const orgId = result[0].ou.org_id;
+      const email = result[0].us.email;
+      this.orgs = result[1];
```
Dimple16, line 212: why are we resetting `this.orgs` here?

Dimple16 (changes requested):

src/app/auth/switch-org/switch-org.page.ts
yashfalke77: now we are refetching the orgs info when the pop up of org not verified is displayed to user. But there can be a case that due to user network connection this can also fails, for the safety point of added this

src/app/auth/switch-org/switch-org.page.ts
yashfalke77: ok

src/app/auth/switch-org/switch-org.page.ts
yashfalke77: handling it in handleDismissPopup function if the data is undefined or null, it will have the default parameter

src/app/auth/switch-org/switch-org.page.ts
yashfalke77: Yes, the problem was due to users network issue  the orgs were not setted up,due to which value of this.orgs was undefined, so to solve this when the user not verified for particular org  popup is displayed we will again fetch it and reset the `this.orgs`

yashfalke77:
> https://sentry.io/organizations/fyle-technologies-private-limi/issues/3840499445/?project=5622998&referrer=release-issue-stream
> 
> How is this issue fixed?

Due to network issue , we were not able get the this.orgs from api call, to solve this whenever the dialogue is displayed we will refetch the org through the api call, even if this fails we are applying safety `?` in `this.orgs?.length`

src/app/auth/switch-org/switch-org.page.ts
yksolanki9: Rephrasing: Why do we need a class member var? We're already getting orgs from the forkjoin, pass it inside the method and we're gtg

src/app/auth/switch-org/switch-org.page.ts
```diff
@@ -204,11 +204,15 @@ export class SwitchOrgPage implements OnInit, AfterViewChecked {
...
+    const eou$ = from(this.authService.getEou());
+    forkJoin({
+      eou: eou$,
+      orgs: this.orgs$,
+    }).subscribe(async ({ eou, orgs }) => {
       const orgName = eou.ou.org_name;
       const orgId = eou.ou.org_id;
       const email = eou.us.email;
-
+      this.orgs = orgs;
```
yksolanki9, line 215: Not needed

yksolanki9 (changes requested):

src/app/auth/switch-org/switch-org.page.ts
yashfalke77: ok

yksolanki9 (approved):

yksolanki9 (commented):
Minor: Remove the unused `orgs` member variable

Dimple16 (approved):

PR #1576, opened 2023-01-03
Code coverage: Org user service changes - 2

Dimple16:
### Changes made

- Added param and method return types
- Added new models
- Added changes to org user spec file
- Added initial 2 test cases

### Current org user service coverage
Screenshot 2023-01-03 at 5 42 03 PM


src/app/core/models/employee-params.model.ts
```diff
@@ -0,0 +1,11 @@
+export interface EmployeeParams {
```
Julias0, line 1: Use partial if everything is optional

src/app/core/models/unflattened-eou.model.ts
```diff
@@ -0,0 +1,61 @@
+export interface UnflattenedEou {
```
Julias0, line 1: `EouApiResponse` instead of `UnflattenedEou`

Julias0 (approved):
Minor comment - fix and merge

PR #1584, opened 2023-01-05
Code coverage:  Org user service - changes 3&4

Dimple16:
Screenshot 2023-01-05 at 4 50 01 PM


src/app/core/services/org-user.service.spec.ts
```diff
@@ -77,4 +90,166 @@ describe('OrgUserService', () => {
...
+    orgUserService.switchToDelegatee().subscribe((res) => {
+      expect(res).toEqual(extendedOrgUserResponse);
+      done();
+    });
+  });
+
+  it('should return false if the user is not switched to a delegator', async () => {
+    jwtHelperService.decodeToken.and.returnValue(accessTokenData);
+    const token =
+      'eyJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE2NzI5MTcyMDAsImlzcyI6IkZ5bGVBcHAiLCJ1c2VyX2lkIjoidXNNakxpYm15ZTdzIiwib3JnX3VzZXJfaWQiOiJvdXJ3N0hpNG1tcE8iLCJvcmdfaWQiOiJvck5WdGhUbzJaeW8iLCJyb2xlcyI6IltcIkZZTEVSXCIsXCJGSU5BTkNFXCIsXCJBRE1JTlwiLFwiQVBQUk9WRVJcIixcIlZFUklGSUVSXCIsXCJQQVlNRU5UX1BST0NFU1NPUlwiLFwiSE9QXCJdIiwic2NvcGVzIjoiW10iLCJhbGxvd2VkX0NJRFJzIjoiW10iLCJ2ZXJzaW9uIjoiMyIsImNsdXN0ZXJfZG9tYWluIjoiXCJodHRwczovL3N0YWdpbmcuZnlsZS50ZWNoXCIiLCJleHAiOjE2NzI5MjA4MDB9.hTMJ56cPH_HgKhZSKNCOIEGAzaAXCfIgbEYcaudhXwk';
```
Julias0, line 231: add a comment on what this token contains

src/app/core/services/org-user.service.spec.ts
```diff
@@ -77,4 +90,166 @@ describe('OrgUserService', () => {
...
+
+    orgUserService.isSwitchedToDelegator().then(async (res) => {
+      expect(res).toEqual(false);
+    });
+  });
+
+  it('should return true if the user is switched to a delegator', async () => {
+    jwtHelperService.decodeToken.and.returnValue(accessTokenWithProxyOrgUserId);
+    const token =
+      'eyJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE2NzI5MTcxNTgsImlzcyI6IkZ5bGVBcHAiLCJ1c2VyX2lkIjoidXNCa0pEMVVtMTc0Iiwib3JnX3VzZXJfaWQiOiJvdTVxclBKYkdmV00iLCJvcmdfaWQiOiJvck5WdGhUbzJaeW8iLCJyb2xlcyI6IltcIkZZTEVSXCIsXCJWRVJJRklFUlwiXSIsInNjb3BlcyI6IltdIiwicHJveHlfb3JnX3VzZXJfaWQiOiJvdXJ3N0hpNG1tcE8iLCJhbGxvd2VkX0NJRFJzIjoiW10iLCJ2ZXJzaW9uIjoiMyIsImNsdXN0ZXJfZG9tYWluIjoiXCJodHRwczovL3N0YWdpbmcuZnlsZS50ZWNoXCIiLCJleHAiOjE2NzI5MjA3NTh9.VqpiTmEd_Kp-fK11gBV-VfjEkPhCja-diu-TGDGPeKA';
```
Julias0, line 245: add a comment on what this token contains

src/app/core/services/org-user.service.spec.ts
```diff
@@ -77,4 +90,166 @@ describe('OrgUserService', () => {
...
+  it('should return false if the user is not switched to a delegator', async () => {
+    jwtHelperService.decodeToken.and.returnValue(accessTokenData);
+    const token =
+      'eyJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE2NzI5MTcyMDAsImlzcyI6IkZ5bGVBcHAiLCJ1c2VyX2lkIjoidXNNakxpYm15ZTdzIiwib3JnX3VzZXJfaWQiOiJvdXJ3N0hpNG1tcE8iLCJvcmdfaWQiOiJvck5WdGhUbzJaeW8iLCJyb2xlcyI6IltcIkZZTEVSXCIsXCJGSU5BTkNFXCIsXCJBRE1JTlwiLFwiQVBQUk9WRVJcIixcIlZFUklGSUVSXCIsXCJQQVlNRU5UX1BST0NFU1NPUlwiLFwiSE9QXCJdIiwic2NvcGVzIjoiW10iLCJhbGxvd2VkX0NJRFJzIjoiW10iLCJ2ZXJzaW9uIjoiMyIsImNsdXN0ZXJfZG9tYWluIjoiXCJodHRwczovL3N0YWdpbmcuZnlsZS50ZWNoXCIiLCJleHAiOjE2NzI5MjA4MDB9.hTMJ56cPH_HgKhZSKNCOIEGAzaAXCfIgbEYcaudhXwk';
+    tokenService.getAccessToken.and.returnValue(Promise.resolve(token));
+
+    const result = await orgUserService.isSwitchedToDelegator();
+    expect(result).toBe(false);
+
+    orgUserService.isSwitchedToDelegator().then(async (res) => {
```
Julias0, line 237: Why are the same tests done twice?

src/app/core/services/org-user.service.spec.ts
```diff
@@ -77,4 +90,166 @@ describe('OrgUserService', () => {
...
+      order: 'us_full_name.asc,us_email.asc,ou_id',
+      us_email: 'in.(ajain@fyle.in)',
+      or: '(ou_status.like.*"ACTIVE",ou_status.like.*"PENDING_DETAILS")',
+      and: '(or(ou_status.like.*"ACTIVE",ou_status.like.*"PENDING_DETAILS"),or(ou_status.like.*"ACTIVE",ou_status.like.*"PENDING_DETAILS"))',
+    };
+    apiV2Service.get.and.returnValue(of(employeesParamsRes));
+    orgUserService.getEmployeesByParams(params).subscribe((res) => {
+      expect(res).toEqual(employeesParamsRes);
+    });
+    expect(apiV2Service.get).toHaveBeenCalledWith('/spender_employees', { params });
```
Julias0, line 135: Move this inside the subscribe otherwise you may hit a race condition

src/app/core/services/org-user.service.spec.ts
```diff
@@ -77,4 +90,166 @@ describe('OrgUserService', () => {
...
+    orgUserService.getEmployeesBySearch(params).subscribe((res) => {
```
Julias0, line 136: Move this to a different test.

src/app/core/services/org-user.service.spec.ts
```diff
@@ -77,4 +90,166 @@ describe('OrgUserService', () => {
...
+      ou_roles: 'like.%ADMIN%',
+      ou_status: 'eq."ACTIVE"',
+      select: 'us_full_name,us_email',
+    };
+    orgUserService.getEmployeesByParams(params).subscribe((res) => {
+      expect(res).toEqual(employeesRes);
+      done();
+    });
+
+    expect(apiV2Service.get).toHaveBeenCalledWith('/spender_employees', { params });
```
Julias0, line 121: Move this inside the subscribe otherwise this may lead to a race condition

Julias0 (changes requested):
Added comments

src/app/core/services/org-user.service.spec.ts
Dimple16: we already have a different test for `getEmployeesByParams`, in the service `getEmployeesBySearch` method, calls the `getEmployeesByParams` to return the value, we need the extra method call for `getEmployeesBySearch`

Julias0: The reason I am asking this is that this can cause a race condition - If this subscribe finishes first, the done method will be called and the test will be passed even though the other subscribe block may execute later and fail.

Dimple16: I've removed the extra subscribe block,

Julias0 (approved):

PR #1586, opened 2023-01-10
 task  FYLE-85zrh6q1u , Audit trails header should be 'Expense Added to Report' 

CODEX108:
@Dimple16 
The Audit header trails **before** :
dd84e442-f599-4715-89a6-b31dadcd4e92_medium

The Updated Audit header trails:

Changed File : ```src\app\core\services\status.service.ts```


Dimple16 (changes requested):
remove the package-lock.json file from PR

Dimple16 (approved):

PR #1587, opened 2023-01-10
task 20d8kmq Create Expense: Mileage expense: Select location pop up.~ There should be some top spacing

CODEX108:
@Dimple16

### DIFF
--- a/src/global.scss
+++ b/src/global.scss
@@ -402,7 +402,7 @@ ion-popover.camera-options-popover {
 
 .mat-form-field-no-padding {
   .mat-form-field-wrapper {
-    padding-top: 12px;
+    padding: 0;
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/global.scss
```diff
@@ -402,7 +402,7 @@ ion-popover.camera-options-popover {
 
 .mat-form-field-no-padding {
   .mat-form-field-wrapper {
-    padding: 0;
+    padding-top: 12px;
```
Dimple16, line 405: this is a global change, this will impact all files using this class, you will have to change it specific to the location pop up

Dimple16 (changes requested):

src/global.scss
CODEX108: ok

Dimple16 (approved):

PR #1591, opened 2023-01-11
Added unit tests for select project

Julias0:


src/app/fyle/add-edit-expense/add-edit-expense.page.html
```diff
@@ -438,6 +438,8 @@
                         [mandatory]="txnFields?.project_id?.is_mandatory"
                         [recentlyUsed]="recentProjects"
                         formControlName="project"
+                        [validInParent]="fg.controls.project.valid"
+                        [touchedInParent]="fg.controls.project.touched"
```
Julias0, line 442: This is required to avoid using injected NgControl inside the component. That was a pre-existing hack

Julias0: It is mandatory for the child to know the touched state of the parent form control because the markAllAsTouched we do when users click on save, doesn't propage through controls

Julias0: Same for valid

src/app/shared/components/fy-select-project/fy-select-project.component.spec.ts
```diff
@@ -1,26 +1,268 @@
...
+      project_org_category_ids: [122269, 122270, 122271, null],
+      project_org_id: 'orFdTTTNcyye',
+      project_updated_at: new Date('2021-07-08T10:28:27.686886'),
+      projectv2_name: 'Customer Mapped Project',
+      sub_project_name: null,
+    };
+
+    component.writeValue(project);
+    fixture.detectChanges();
+    await fixture.whenStable();
```
rvab, line 85: * Why use `fixture.whenStable()`? Is there any promise/setTimeout involved?
* Can we use `fakeAsync` instead of `async` and use `flushMicroTasks`?

src/app/shared/components/fy-select-project/fy-select-project.component.spec.ts
```diff
@@ -1,26 +1,268 @@
...
+import { ModalPropertiesService } from 'src/app/core/services/modal-properties.service';
+import { CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
 
 import { FySelectProjectComponent } from './fy-select-project.component';
 
-xdescribe('FySelectProjectComponent', () => {
+describe('FySelectProjectComponent', () => {
   let component: FySelectProjectComponent;
   let fixture: ComponentFixture<FySelectProjectComponent>;
+  let el: DebugElement;
```
rvab, line 14: `el` -> `element`

src/app/shared/components/fy-select-project/fy-select-project.component.spec.ts
```diff
@@ -1,26 +1,268 @@
...
   beforeEach(
     waitForAsync(() => {
+      const modalControllerSpy = jasmine.createSpyObj('ModalController', ['create']);
+      const modalPropertiesServiceSpy = jasmine.createSpyObj('ModalPropertiesService', ['getModalDefaultProperties']);
+
       TestBed.configureTestingModule({
         declarations: [FySelectProjectComponent],
-        imports: [IonicModule.forRoot()],
+        imports: [IonicModule.forRoot(), FormsModule, ReactiveFormsModule],
+        schemas: [CUSTOM_ELEMENTS_SCHEMA],
```
rvab, line 26: why is this required?

src/app/shared/components/fy-select-project/fy-select-project.component.spec.ts
```diff
@@ -1,26 +1,268 @@
...
+    component.writeValue(null);
+    fixture.detectChanges();
+    await fixture.whenStable();
+
+    const inputDisplayField = el.query(By.css('.fy-select-project--input'));
+    expect(inputDisplayField.nativeElement.value).toEqual('');
+  });
+
+  it('should be able to set touched value properly whenever the input field is touched', () => {
+    const callBackTestObject = jasmine.createSpyObj('touchedTested', ['touchedTested']);
```
rvab, line 101: is `touchedTested` a dummy callback function?

src/app/shared/components/fy-select-project/fy-select-project.component.spec.ts
```diff
@@ -1,26 +1,268 @@
...
+
+  it('should be able to display proper label', () => {
+    component.label = 'Project Field';
+    fixture.detectChanges();
+    const label = el.query(By.css('.fy-select-project--label-content'));
+    expect(label.nativeElement.textContent.trim()).toEqual('Project Field');
+  });
+
+  it('should be able to set a proper project value', async () => {
+    const project = {
```
rvab, line 63: Create a model and test data file for this and import it here.

src/app/fyle/add-edit-expense/add-edit-expense.page.html
rvab: > This is required to avoid using injected NgControl inside the component. That was a pre-existing hack and that hack would cause getting code coverage in certain places impossible

Didn't understand the comment. Angular Material is using `ngControl`. https://github.com/angular/components/blob/main/src/material/input/input.ts. Why can't we use it?

rvab (changes requested):

src/app/fyle/add-edit-expense/add-edit-expense.page.html
Julias0: ngControl is not directly injected into this component. If you looked at the material component, the ngControl is directly injected. 
We cant directly inject NgControl into this component since it causes a circular dependency error, hence I added a hack to do it using injector. But dependency injection using injector makes this almost impossible to unit test.

Just removing this and replacing it with inputs makes it much much more easier to write the tests, circumventing the whole problem.

src/app/shared/components/fy-select-project/fy-select-project.component.spec.ts
Julias0: This is required to prevent ionic components which are being used in the template from throwing errors.

Ref: https://stackoverflow.com/questions/39428132/custom-elements-schema-added-to-ngmodule-schemas-still-showing-error

src/app/shared/components/fy-select-project/fy-select-project.component.spec.ts
Julias0: Without fixture.whenStable(), the changes were not reflecting inside the DOM component value property which would cause the assertion to fail. My assumption is, the change detection needs some time for the ngModel value change to take effect.

The only async task here is waiting for the fixture to get stable. Since the async operation is not really inside the component, I feel fakeAsync/flushMicroTasks is overkill

src/app/shared/components/fy-select-project/fy-select-project.component.spec.ts
Julias0: Yes. Basically I will check if this method has been called if the blur event gets triggered in the input field. This gives me a clean way to do this

src/app/shared/components/fy-select-project/fy-select-project.component.spec.ts
Julias0: I find this easier to read rather than jumping between multiple files. This was meant to serve as an example to work off rather than be an actual PR I will merge in.

rvab (approved):

PR #1597, opened 2023-01-13
Removed loader inside camera preview

Julias0:


yksolanki9 (approved):
Can remove import of LoaderService as it's unused now

PR #1601, opened 2023-01-13
Code coverage: Expense fields service - statements 28.84%

Dimple16:
Screenshot 2023-01-17 at 6 28 34 PM


src/app/core/mock-data/expense-field.data.ts
```diff
@@ -0,0 +1,1410 @@
+import { ExpenseField } from '../models/v1/expense-field.model';
+
+export const expenseFieldResponse: ExpenseField[] = [
+  {
+    id: 2,
+    created_at: new Date('218-01-31T23:50:27.221Z'),
```
Julias0, line 6: apiService dosnt return the dates inside a new Date object.This is missing functionality that should be handled inside the service

src/app/core/mock-data/expense-fields-map.data.ts
```diff
@@ -0,0 +1,1894 @@
+import { ExpenseFieldsMap } from '../models/v1/expense-fields-map.model';
+
+export const txnFieldsData: Partial<ExpenseFieldsMap> = {
+  purpose: {
```
Julias0, line 4: again, this object is too huge

src/app/core/models/v1/expense-fields-map.model.ts
```diff
@@ -1,84 +1,84 @@
...
-  timestamp_column5: ExpenseField[];
-  to_dt: ExpenseField[];
-  train_travel_class: ExpenseField[];
-  txn_dt: ExpenseField[];
-  vendor_id: ExpenseField[];
-  billable: ExpenseField[];
-  tax_group_id: ExpenseField[];
-  project_id: ExpenseField[];
-  org_category_id: ExpenseField[];
+  boolean_column1: ExpenseField[] | ExpenseField;
```
Julias0, line 4: In what case is this an array instead of a single expense field? This model is very weird - we should refactor and remove the dependency on either the array or the single value

Julias0 (changes requested):
Added comments

PR #1607, opened 2023-01-17
Code Coverage: Dashboard Service #3

jayfyle:
# Subtask #6
Almost 100% code coverage across the board, only a few branches left
Almost_100_code_coverage_dashboard

## Edit 1: Code Coverage now 100% including branches

Dashboard_100_coverage


src/app/core/mock-data/card-aggregate-stat.data.ts
```diff
@@ -0,0 +1,186 @@
+import { CardAggregateStat } from '../models/card-aggregate-stat.model';
+
+export const cardAggregateStatParam: CardAggregateStat[] = [
+  {
```
Dimple16, line 4: let's reduce this list to have 3 items?

src/app/core/mock-data/ccc-expense.details.data.ts
```diff
@@ -0,0 +1,190 @@
+import { CCCDetails } from '../models/ccc-expense-details.model';
+
+export const expectedAssignedCCCStats: CCCDetails = {
+  totalTxns: 1090,
+  totalAmount: 589553.832,
+  cardDetails: [
+    {
+      aggregates: [
```
Dimple16, line 8: for any list, let's keep 2/3 items max

src/app/core/mock-data/extended-org-user.data.ts
```diff
@@ -0,0 +1,163 @@
...
+    branch_ifsc: 'ICIC0002322',
+    branch_account: null,
+    mobile: '123456',
+    mobile_verified: false,
+    mobile_verified_at: null,
+    is_primary: true,
+    owner: false,
+    joining_dt: new Date('2017-07-25T00:00:00.000+0000'),
+    special_email: 'receipts+ajain_6@fyle.ai',
+    custom_field_values: [
```
Dimple16, line 38: keep 1/2 custom values to compare

src/app/core/mock-data/stats-response.data.ts
```diff
@@ -141,6 +134,262 @@ export const apiReportStatsRes: StatsResponse = {
...
+            },
+          ],
+        },
+      ],
+    },
+  ],
+  url: '/v2/reports/stats',
+};
+
+export const apiAssignedCardDetailsRes: StatsResponse = {
```
Dimple16, line 185: this data is huge too, you can reduce it

src/app/core/mock-data/stats.data.ts
```diff
@@ -9,3 +9,8 @@ export const expectedIncompleteExpStats: Stats = {
   count: 1130,
   sum: 1148487.57555147,
 };
+
+export const expectedEmptyStats: Stats = {
+  count: undefined,
```
Dimple16, line 14: is this to cover any branch? this will never be the result no? or should it be null?

src/app/core/mock-data/unique-cards-stats.data.ts
```diff
@@ -0,0 +1,54 @@
+import { UniqueCardStats } from '../models/unique-cards-stats.model';
+
+export const expectedUniqueCardStats: UniqueCardStats[] = [
```
Dimple16, line 3: reduce

src/app/core/mock-data/unique-cards.data.ts
```diff
@@ -0,0 +1,32 @@
+import { UniqueCards } from '../models/unique-cards.model';
+
+export const uniqueCardsParam: UniqueCards[] = [
+  {
```
Dimple16, line 4: reduce

src/app/fyle/dashboard/dashboard.service.spec.ts
```diff
@@ -69,6 +95,32 @@ describe('DashboardService', () => {
...
+      expect(res).toEqual(expectedEmptyStats);
+      expect(transactionService.getTransactionStats).toHaveBeenCalledWith(
+        'count(tx_id),sum(tx_amount)',
+        apiUnreportedParams
+      );
+      done();
+    });
+  });
+
+  it('getUnreportedExpensesStats():  should return null INCOMPLETE expense stats as api sends empty values as response', (done) => {
```
Dimple16, line 111: `should return empty`?

Dimple16 (changes requested):
added comments

src/app/core/mock-data/stats.data.ts
jayfyle: When there is no relevant data, the function sends undefined as values

jayfyle: It is used in branch coverage, for edge cases when the API does not send any data or the aggregate value be an empty list. In all such cases, the object returned has sum and count as undefined.

Dimple16 (approved):

src/app/fyle/dashboard/dashboard.service.spec.ts
```diff
@@ -69,6 +95,32 @@ describe('DashboardService', () => {
     });
   });
 
+  it('getUnreportedExpensesStats(): should return empty or undefined UNREPORTED expense stats as api sends empty list for aggregates', (done) => {
+    transactionService.getTransactionStats.and.returnValue(of(apiTxnUnreportedStatsEmptyRes));
+
+    dashboardService.getUnreportedExpensesStats().subscribe((res) => {
+      expect(res).toEqual(expectedEmptyStats);
+      expect(transactionService.getTransactionStats).toHaveBeenCalledWith(
```
rvab, line 103: don't we have `toHaveBeenCalledTImes(number)` in jasmine?

src/app/fyle/dashboard/dashboard.service.spec.ts
```diff
@@ -69,6 +95,32 @@ describe('DashboardService', () => {
     });
   });
 
+  it('getUnreportedExpensesStats(): should return empty or undefined UNREPORTED expense stats as api sends empty list for aggregates', (done) => {
```
rvab, line 98: Is the test description correct? is it `should return empty or undefined for UNREPORTED expense` or `should return empty or undefined UNREPORTED expense`?

src/app/core/mock-data/stats-response.data.ts
```diff
@@ -1,20 +1,13 @@
+import { Aggregate, Value, Key } from '../models/v2/stats-response.model';
+
 type StatsResponse = {
```
rvab, line 3: Do we need a new `type`? Can this be done with `interface`?  Is this being used anywhere else?

src/app/core/models/stats-dimension-response.model.ts
```diff
@@ -1,5 +1,5 @@
 export interface StatsDimensionResponse {
-  aggregates: {
+  aggregates?: {
```
rvab, line 2: Why make it optional?

rvab (changes requested):

src/app/core/models/stats-dimension-response.model.ts
jayfyle: I was trying to increase branch coverage, and certain branches required aggregates not being present. Hence I made it optional. Here's a screenshots for reference:

branches_to_be_covered

Still_remaining_branches

src/app/core/mock-data/stats-response.data.ts
jayfyle: The original type has been written as a class, so it was throwing errors while trying to explicitly cast it due to the member functions. To avoid initialising it the conventional way, I created an additional type of the same name and structure, only to cast the responses in the expected type. A similar implementation can be found in task service as well.

src/app/fyle/dashboard/dashboard.service.spec.ts
jayfyle: Sorry, my bad. It is a typo, meant to write the first one. I have fixed it for all such cases

src/app/fyle/dashboard/dashboard.service.spec.ts
jayfyle: Added it for all cases

rvab (approved):

PR #1610, opened 2023-01-17
Code coverage: Expense fields changes - 6

Dimple16:
Screenshot 2023-01-19 at 3 19 14 PM


src/app/core/services/expense-fields.service.spec.ts
```diff
@@ -54,4 +60,28 @@ describe('ExpenseFieldsService', () => {
...
+  });
+
+  it('getAllMap(): should get all expense fields map', (done) => {
+    authService.getEou.and.returnValue(new Promise((resolve) => resolve(extendedOrgUserResponse)));
+    apiService.get.and.returnValue(of(expenseFieldWithSeq));
+
+    expenseFieldsService.getAllMap().subscribe((expenseFieldsMap) => {
+      expect(expenseFieldsMap).toEqual(expenseFieldsMapResponse2);
+    });
+    done();
```
rvab, line 76: this should inside subscribe no?

src/app/core/services/expense-fields.service.spec.ts
```diff
@@ -54,4 +60,28 @@ describe('ExpenseFieldsService', () => {
...
+  });
+
+  it('filterbyCategory(): should filter expense fields by category', () => {
+    const fields = ['purpose', 'txn_dt', 'vendor_id', 'org_category_id'];
+    expenseFieldsService
+      .filterByOrgCategoryId(expenseFieldsMapResponse, fields, orgCategoryData)
+      .subscribe((expenseFields) => {
+        expect(expenseFields).toEqual(expenseFieldObjData);
```
rvab, line 84: `done` is not called?

rvab (changes requested):

src/app/core/services/expense-fields.service.ts
```diff
@@ -26,7 +28,8 @@ export class ExpenseFieldsService {
             is_custom: false,
           },
         })
-      )
+      ),
+      map((res) => this.dateService.fixDates(res))
```
rvab, line 32: Why is this added?

src/app/core/services/expense-fields.service.ts
```diff
@@ -68,7 +71,7 @@ export class ExpenseFieldsService {
     tfcMap: Partial<ExpenseFieldsMap>,
     fields: string[],
     orgCategory: OrgCategory
-  ): Observable<Partial<ExpenseFieldsMap>> {
+  ): Observable<Partial<ExpenseFieldsMap | ExpenseFieldsObj>> {
```
rvab, line 74: Why create a new model?

src/app/core/services/expense-fields.service.ts
```diff
@@ -113,7 +116,9 @@ export class ExpenseFieldsService {
     );
   }
 
-  getDefaultTxnFieldValues(txnFields: Partial<ExpenseFieldsMap>): DefaultTxnFieldValues {
+  getDefaultTxnFieldValues(
+    txnFields: Partial<ExpenseFieldsMap> | Partial<ExpenseFieldsObj>
```
rvab, line 120: same as above

rvab (approved):
Minor comments

src/app/core/services/expense-fields.service.ts
Dimple16: since the service is not handling the conversion of date to string, we need this as the DB and model type is `Date`, but the value is of string,

PR #1611, opened 2023-01-18
Create modela and mock service for dependent fields

yksolanki9:


src/app/core/models/dependent-field.model.ts
```diff
@@ -0,0 +1,8 @@
+export interface DependentField {
+  id: number;
+  name: string;
+  is_enabled: boolean;
+  is_mandatory: boolean;
+  placeholder: string;
+  dependent_field_id: number;
```
Julias0, line 7: Shouldnt this be parent field id?

src/app/core/services/dependent-fields.service.ts
```diff
@@ -0,0 +1,83 @@
...
+import { DependentField } from '../models/dependent-field.model';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class DependentFieldsService {
+  constructor() {}
+
+  getDependentFields(): Observable<DependentField[]> {
+    return of([
```
Julias0, line 13: add a few samples of non mandatory dependant field and a few cases of disabled dependant fields too

src/app/core/services/dependent-fields.service.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  generateOptions(fieldId: number, parentFieldId: number) {
+    const val = [];
+    for (let i = 0; i <= 10; i++) {
+      val.push({
+        id: fieldId * i,
+        name: `Parent field ${fieldId}, Option ${i}`,
+        is_enabled: true,
+        field_id: parentFieldId,
```
Julias0, line 64: Shouldnt this be equal to fieldId? 
The parent field id is the id of the parent right - so parentFieldId and parentValueId define which values would come up

Julias0 (changes requested):
Added comments

src/app/core/models/dependent-field.model.ts
yksolanki9: Every field will have the dependent field of its child which makes it easier to get the child field.
If we use parent field id here, then we’ll have to loop through all the fields and match the parent_field_id key to get the dependent field

src/app/core/services/dependent-fields.service.ts
yksolanki9: Yes, have fixed this

src/app/core/services/dependent-fields.service.ts
yksolanki9: I've made some fields non mandatory and have disabled some others. 
But, while developing, I'll be changing this settings everytime. Else will have to build custom dependency chains for every case which is not needed

Julias0: Works for me 👍

Julias0 (approved):

PR #1615, opened 2023-01-19
Code coverage: Report Service #2

jayfyle:
Removed unused code from report service and added params and return types for some functions

### DIFF
--- a/src/app/core/services/report.service.ts
+++ b/src/app/core/services/report.service.ts
@@ -13,13 +13,16 @@ import { isEqual } from 'lodash';
 import { DataTransformService } from './data-transform.service';
 import { Cacheable, CacheBuster } from 'ts-cacheable';
 import { TransactionService } from './transaction.service';
-import { StatsResponse } from '../models/v2/stats-response.model';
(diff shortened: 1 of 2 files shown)


src/app/core/services/report.service.ts
```diff
@@ -96,7 +99,7 @@ export class ReportService {
   @Cacheable({
     cacheBusterObserver: reportsCacheBuster$,
   })
-  getERpt(rptId) {
+  getERpt(rptId: string) {
```
Dimple16, line 102: return type?

src/app/core/services/report.service.ts
```diff
@@ -112,7 +115,7 @@ export class ReportService {
   @CacheBuster({
     cacheBusterNotifier: reportsCacheBuster$,
   })
-  addTransactions(rptId, txnIds) {
+  addTransactions(rptId: string, txnIds: string[]) {
```
Dimple16, line 118: return type?

Dimple16 (approved):
many return types are missing, are we handling it in another PR?

PR #1617, opened 2023-01-20
Code coverage: Transaction service - Subtask-3

Dimple16:
### Changes made
- Added missing models
- Added param and return types for all methods

### DIFF
--- a/src/app/core/models/sort-filters-params.model.ts
+++ b/src/app/core/models/sort-filters-params.model.ts
@@ -1,0 +1,9 @@
+import { FilterQueryParams } from './filter-query-params.model';
+
+export interface SortFiltersParams {
+  pageNumber: number;
+  queryParams: FilterQueryParams;
+  sortParam: string;
+  sortDir: string;
+  searchString: string;
+}
--- a/src/app/core/services/transaction.service.ts
+++ b/src/app/core/services/transaction.service.ts
@@ -27,6 +27,14 @@ import { AccountsService } from './accounts.service';
 import { SpenderPlatformApiService } from './spender-platform-api.service';
 import { PlatformPolicyExpense } from '../models/platform/platform-policy-expense.model';
 import { ExpensePolicy } from '../models/platform/platform-expense-policy.model';
+import { EtxnParams } from '../models/etxn-params.model';
+import { ApiV2Response } from '../models/v2/api-v2-response.model';
+import { Transaction } from '../models/v1/transaction.model';
+import { FileObject } from '../models/file_obj.model';
+import { UnflattenedTransaction } from '../models/unflattened-transaction.model';
+import { CurrencySummary } from '../models/currency-summary.model';
+import { FilterQueryParams } from '../models/filter-query-params.model';
+import { SortFiltersParams } from '../models/sort-filters-params.model';
 
 enum FilterState {
   READY_TO_REPORT = 'READY_TO_REPORT',
@@ -40,6 +48,15 @@ const transactionsCacheBuster$ = new Subject<void>();
 type PaymentMode = {
   name: string;
   key: string;
+};
+
+type PaymentModeSummary = {
+  [paymentMode: string]: {
+    name: string;
+    key: string;
+    amount: number;
+    count: number;
+  };
 };
 
 @Injectable({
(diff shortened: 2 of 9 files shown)


src/app/core/models/sort-filters-params.model.ts
```diff
@@ -0,0 +1,9 @@
+import { FilterQueryParams } from './filter-query-params.model';
+
+export interface SortFiltersParams {
+  pageNumber: number;
+  queryParams: FilterQueryParams;
+  sortParam: string;
+  sortDir: string;
```
rvab, line 7: What will be the values here?. If it's `desc` and `asc`, then you can use `sortDir: 'asc' | 'desc'`. This helps in preventing passing random string

src/app/core/services/transaction.service.ts
```diff
@@ -42,6 +50,15 @@ type PaymentMode = {
   key: string;
 };
 
+type PaymentModeSummary = {
```
rvab, line 53: We are creating models/types in different files, no? Any reason for keeping it here?

src/app/core/services/transaction.service.ts
```diff
@@ -136,12 +153,12 @@ export class TransactionService {
     cacheBusterObserver: transactionsCacheBuster$,
   })
   getMyExpenses(
-    config: Partial<{ offset: number; limit: number; order: string; queryParams: any }> = {
+    config: Partial<{ offset: number; limit: number; order: string; queryParams: {} }> = {
```
rvab, line 156: * what is the type of `queryParams`? If we can't define model for it, let's use [`Record`](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type)

src/app/core/services/transaction.service.ts
```diff
@@ -174,7 +191,7 @@ export class TransactionService {
   @Cacheable({
     cacheBusterObserver: transactionsCacheBuster$,
   })
-  getAllExpenses(config: Partial<{ order: string; queryParams: any }>) {
+  getAllExpenses(config: Partial<{ order: string; queryParams: {} }>): Observable<Expense[]> {
```
rvab, line 194: same as above

src/app/core/services/transaction.service.ts
```diff
@@ -189,14 +206,14 @@ export class TransactionService {
...
-      reduce((acc, curr) => acc.concat(curr), [] as any[])
+      reduce((acc, curr) => acc.concat(curr), [] as Expense[])
     );
   }
 
   @Cacheable({
     cacheBusterObserver: transactionsCacheBuster$,
   })
-  getTransactionStats(aggregates: string, queryParams = {}) {
+  getTransactionStats(aggregates: string, queryParams = {}): Observable<any> {
```
rvab, line 216: * same comment as above for `queryParams`.
* Why use `any` here?

src/app/core/services/transaction.service.ts
```diff
@@ -363,23 +380,23 @@ export class TransactionService {
     );
   }
 
-  getETxnc(params: { offset: number; limit: number; params: any }) {
+  getETxnc(params: { offset: number; limit: number; params: {} }): Observable<Expense[]> {
```
rvab, line 383: same comment as above for queryParams

src/app/core/services/transaction.service.ts
```diff
@@ -363,23 +380,23 @@ export class TransactionService {
...
     return this.apiV2Service
       .get('/expenses', {
         ...params,
       })
       .pipe(map((etxns) => etxns.data));
   }
 
-  getMyExpensesCount(queryParams = {}) {
+  getMyExpensesCount(queryParams = {}): Observable<number> {
```
rvab, line 391: same as above

rvab (changes requested):

src/app/core/services/transaction.service.ts
Dimple16: - fixed all the references where `queryParams` was used
- the usage of `getTransactionStats` needs to be fixed throughout, in other files as well, a lot of changes will get involved, added a `TODO` here

src/app/core/services/transaction.service.ts
Dimple16: this is moved to a different file

src/app/core/models/sort-filters-params.model.ts
Dimple16: the usage of `sortDir` in the sort method is different than how it is used in the Filters model, we will need to change it throughout, so using the same here, in few places an enum is used, it is incompatible here

rvab (approved):

PR #1618, opened 2023-01-23
Dependent Fields: Create new component for dependent field

yksolanki9:
Will remove ngControl, getter and setter after creating the modal component and will use formControl for setting the value the field instead of ngModel


src/app/shared/components/dependent-field/dependent-field.component.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  private onChangeCallback: (_: any) => void = noop;
+
+  constructor(
+    private modalController: ModalController,
+    private injector: Injector,
+    private modalProperties: ModalPropertiesService
+  ) {}
+
+  get valid() {
```
yksolanki9, line 54: Will remove this ngControl, getter and setter after creating the modal component and will use formControl for setting the value of this field instead of ngModel

src/app/shared/components/dependent-field/dependent-field.component.html
```diff
@@ -0,0 +1,26 @@
...
+    <span *ngIf="mandatory" class="dependent-field--mandatory"> * </span>
+  </div>
+  <div
+    class="dependent-field--input-container"
+    [ngClass]="{ 'dependent-field--input-invalid': !valid, 'dependent-field--disabled': disabled }"
+  >
+    <input
+      class="dependent-field--input smartlook-show"
+      type="text"
+      [placeholder]="!disabled ? placeholder || 'Select ' + label : ''"
```
Julias0, line 13: why is placeholder linked to disabled

src/app/shared/components/dependent-field/dependent-field.component.ts
Julias0: From the parent you can pass in the valid state and touched state as inputs which can be in turn used for this

src/app/shared/components/dependent-field/dependent-field.component.html
```diff
@@ -0,0 +1,26 @@
...
+      (click)="openModal()"
+      [ngClass]="{ 'dependent-field--input-disabled': disabled }"
+      [disabled]="disabled"
+    />
+    <mat-icon
+      class="dependent-field--icon"
+      [ngClass]="{ 'dependent-field--input-disabled': disabled }"
+      svgIcon="fy-arrow-down"
+    ></mat-icon>
+  </div>
```
Julias0, line 25: What about validation messages? Shouldnt we add them inside this? Or will they be added outside

Julias0 (changes requested):
minor comments and questions

src/app/shared/components/dependent-field/dependent-field.component.ts
yksolanki9: Okay, will handle it in the next PR as discussed

src/app/shared/components/dependent-field/dependent-field.component.html
yksolanki9: Inititally, I thought of showing it from expense form, but since we're anyways passing the validation state to child, we can show this in the child.

yksolanki9: This is done

Julias0 (approved):

PR #1621, opened 2023-01-23
 FYLE-21fmba1 Removed the resubmit report pop up

CODEX108:


Dimple16 (changes requested):
remove all the files related to `ResubmitReportPopoverComponent` once you check if they aren't used anywhere

CODEX108:
> remove all the files related to `ResubmitReportPopoverComponent` once you check if they aren't used anywhere

@Dimple16  It is still used for submitting the report since the previous method still exists.Therefore thought of removing the files after adding the `submitReport()` method.

Dimple16 (approved):

PR #1630, opened 2023-01-25
Code Coverage subtask 85zrja8dj : Snackbar properties service

CODEX108:

### DIFF
--- a/src/app/core/services/snackbar-properties.service.spec.ts
+++ b/src/app/core/services/snackbar-properties.service.spec.ts
@@ -37,13 +37,13 @@ describe('SnackbarPropertiesService', () => {
     expect(properties.duration).toEqual(3000);
   });
 
-  it('should return the correct toastMessageData', () => {
+  it('should return redirection text in data', () => {
     const properties = service.setSnackbarProperties('success', {
       message: 'Success message',
-      redirectiontext: 'go to main page',
+      redirectiontext: 'redirection',
     });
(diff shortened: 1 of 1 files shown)


src/app/core/services/snackbar-properties.service.spec.ts
```diff
@@ -2,15 +2,52 @@ import { TestBed } from '@angular/core/testing';
...
+    const properties = service.setSnackbarProperties('information', { message: 'Information message' });
+    expect(properties.data.icon).toEqual('');
+  });
+
+  it('should return the correct duration', () => {
+    const properties = service.setSnackbarProperties('success', { message: 'Success message' });
+    expect(properties.duration).toEqual(3000);
+  });
+
+  it('should return the correct toastMessageData', () => {
```
jayfyle, line 40: maybe it can be renamed as `should return redirection text in data`?

jayfyle (approved):
A small change in naming a spec. Rest looks good.

src/app/core/services/snackbar-properties.service.spec.ts
CODEX108: > maybe it can be renamed as `should return redirection text in data`? Or mention `redirection` in the spec name.

ok

CODEX108: Changes are done

PR #1642, opened 2023-02-02
Removing dead code - part 1

Dimple16:


src/app/core/services/advance.service.ts
```diff
@@ -77,7 +77,7 @@ export class AdvanceService {
     }).pipe(map((advances) => advances.count));
   }
 
-  fixDates(data: ExtendedAdvance) {
+  private fixDates(data: ExtendedAdvance) {
```
rvab, line 80: don't we have a method in `dateService` to fix dates??

rvab (approved):
minor comment

src/app/core/services/advance.service.ts
Dimple16: we do, this method is specific to advance dates

PR #1656, opened 2023-02-07
Dependent Fields: Show dependent fields in view expense/mileage/per-diem page

yksolanki9:


src/app/core/services/custom-inputs.service.ts
```diff
@@ -41,7 +42,7 @@ export class CustomInputsService {
         customInputs.map((customInput) => ({
           ...customInput,
           parent_field_id: customInput.id === 218227 ? 214662 : customInput.id - 1,
-          type: 'DEPENDENT_SELECT',
+          type: customInput.field_name.length === 3 ? 'DEPENDENT_SELECT' : customInput.type,
```
yksolanki9, line 45: Ignore, dummy data for testing. Will be removed once APIs are available
Have added a todo before the method

Julias0 (approved):

PR #1661, opened 2023-02-10
Unit Test Fix for Report Service

jayfyle:


src/app/core/services/report.service.spec.ts
```diff
@@ -747,6 +744,46 @@ describe('ReportService', () => {
...
+        },
+      };
+
+      const getMyReportsParam1 = {
+        offset: 0,
+        limit: 2,
+        queryParams: {
+          rp_state: 'in.(DRAFT,APPROVER_PENDING,APPROVER_INQUIRY)',
+        },
+        order: undefined,
```
rvab, line 762: if it's `undefined`, then do you still need to pass it?

src/app/core/services/report.service.spec.ts
```diff
@@ -747,6 +744,46 @@ describe('ReportService', () => {
...
+      };
+
+      const getMyReportsParam2 = {
+        offset: 2,
+        limit: 2,
+        queryParams: {
+          rp_state: 'in.(DRAFT,APPROVER_PENDING,APPROVER_INQUIRY)',
+        },
+        order: undefined,
```
rvab, line 771: same as above

rvab (approved):
One minor doubt

src/app/core/services/report.service.spec.ts
jayfyle: Yes, otherwise the test fails, it is an expected property in the object

Dimple16 (approved):

PR #1675, opened 2023-02-16
Code coverage: Org service - subtask3

Dimple16:
Screenshot 2023-02-16 at 7 45 22 PM


src/app/core/mock-data/org.data.ts
```diff
@@ -0,0 +1,24 @@
+import { Org } from '../models/org.model';
+
+export const orgData1: Org[] = [
```
rvab, line 3: it's better to create 2 different variables. One for the current org and another for the primary org.

rvab (approved):
One minor comment

PR #1690, opened 2023-02-20
Code coverage: Custom fields service ~100% 

Dimple16:
Screenshot 2023-02-20 at 5 03 12 PM

Note: Added a `TODO` for the skipped lines


src/app/core/mock-data/custom-property.data.ts
```diff
@@ -0,0 +1,36 @@
+import { CustomProperty } from '../models/custom-properties.model';
+
+export const customPropertiesData: CustomProperty<any>[] = [
```
rvab, line 3: Why `any`?

rvab (approved):

src/app/core/mock-data/custom-property.data.ts
Dimple16: because the property value can be of `any` type, string / string[], Date, number, etc etc

PR #1697, opened 2023-02-20
Code coverage: AdvanceActions Component - 100%

Dimple16:
Screenshot 2023-02-20 at 9 44 43 PM


src/app/fyle/view-team-advance-request/advance-actions/advance-actions.component.spec.ts
```diff
@@ -1,26 +1,122 @@
...
+        can_inquire: true,
+        can_reject: false,
+      };
+      fixture.detectChanges();
+      const inquireButton = getElementBySelector(fixture, '.advance-action--action-head');
+      expect(getTextContent(inquireButton)).toContain('Send Back Advance');
+    });
+
+    it('should show all 3 advance action buttons if all can_approve, can_inquire and can_reject are true', () => {
+      component.actions = {
```
rvab, line 75: * Are these actions exclusive?. If one is true, the other 2 will be false all the time?
* Is there a situation where `can_approve` and `can_inquire` are true and `can_reject` is false?. What is the UI state here?. Shouldn't we cover all cases here?
* what happens if all 3 are false?

src/app/fyle/view-team-advance-request/advance-actions/advance-actions.component.spec.ts
```diff
@@ -1,26 +1,122 @@
...
+    });
+  });
+
+  describe('openAnotherPopover():', () => {
+    it('should dismiss popover with the sendBackAdvance command when send back action button is clicked', () => {
+      const mockCommand = 'sendBackAdvance';
+      // @ts-ignore
+      component.popoverController.dismiss.and.returnValue(Promise.resolve());
+      const sendBackButton = getElementBySelector(fixture, '.advance-action--action') as HTMLElement;
+      sendBackButton.click();
```
rvab, line 94: create a utility function for click https://angular.io/guide/testing-components-scenarios#click-helper

src/app/fyle/view-team-advance-request/advance-actions/advance-actions.component.spec.ts
```diff
@@ -1,26 +1,122 @@
...
+      // @ts-ignore
+      component.popoverController.dismiss.and.returnValue(Promise.resolve());
+      const approveButton = getElementBySelector(fixture, '.advance-action--action') as HTMLElement;
+      approveButton.click();
+      component.openAnotherPopover(mockCommand);
+      // @ts-ignore
+      expect(component.popoverController.dismiss).toHaveBeenCalledWith({ command: mockCommand });
+    });
+
+    it('should dismiss popover with the rejectAdvance command when reject advance action button is clicked', () => {
```
rvab, line 111: same as above

rvab (changes requested):

src/app/fyle/view-team-advance-request/advance-actions/advance-actions.component.spec.ts
Dimple16: > Are these actions exclusive?. If one is true, the other 2 will be false all the time?
- No, not always, 

> Is there a situation where can_approve and can_inquire are true and can_reject is false?. What is the UI state here?. Shouldn't we cover all cases here?
- I have covered usual cases, 
- Most cases for an approver:
      - all 3 buttons will be visible when the advance request isn't approved yet, 
      - once approved, it can only be sent back, so only 1 button sent back will be present

> what happens if all 3 are false?
- The entire advance action button itself isn't visible, added a test case for that too

src/app/fyle/view-team-advance-request/advance-actions/advance-actions.component.spec.ts
Dimple16: added helper method for `click()`

rvab (approved):

PR #1712, opened 2023-02-23
Code coverage: Disabled page component - 100%

Dimple16:
Screenshot 2023-02-23 at 8 12 01 PM

### DIFF
--- a/src/app/auth/disabled/disabled.page.spec.ts
+++ b/src/app/auth/disabled/disabled.page.spec.ts
@@ -34,15 +34,9 @@ describe('DisabledPage', () => {
     expect(component).toBeTruthy();
   });
 
-  it('should call logout on userEventService when onGotoSignInClick is called', () => {
+  it('should logout and navigate to /auth/sign_in when onGotoSignInClick is called', () => {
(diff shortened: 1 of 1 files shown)


src/app/auth/disabled/disabled.page.spec.ts
```diff
@@ -1,26 +1,48 @@
...
     expect(component).toBeTruthy();
   });
+
+  it('should call logout on userEventService when onGotoSignInClick is called', () => {
+    component.onGotoSignInClick();
+
+    expect(userEventService.logout).toHaveBeenCalledTimes(1);
+  });
+
+  it('should navigate to /auth/sign_in when onGotoSignInClick is called', () => {
```
rvab, line 43: both can be combined into a single test case.

rvab (changes requested):

src/app/auth/disabled/disabled.page.spec.ts
Dimple16: this is fixed

rvab (approved):

PR #1725, opened 2023-02-27
Code coverage: FyCategoryIcon Component - 100%

Dimple16:
Screenshot 2023-02-27 at 11 29 39 PM

Note: I haven't added any specific test case for template as such since this is only display of icons


rvab (changes requested):
It's better to skip this and revisit it later?

rvab (approved):

PR #1730, opened 2023-02-28
Dependent Fields: Fix bugs in implementation

yksolanki9:
Fixes two bugs found during testing
- When changing the project to None, the dependent fields are still visible
- When loading form in edit expense, the if there is a child dependent field, but the value selected in parent field does not have any values for the child field, the child field is still shown.

### DIFF
--- a/src/app/fyle/add-edit-expense/add-edit-expense.page.ts
+++ b/src/app/fyle/add-edit-expense/add-edit-expense.page.ts
@@ -4364,16 +4364,22 @@ export class AddEditExpensePage implements OnInit {
       const dependentFieldValue = txCustomProperties.find(
         (customProp) => customProp.name === dependentField.field_name
       );
-      //Add dependent field with selected value
-      this.addDependentField(dependentField, parentField.value, dependentFieldValue?.value);
 
-      //Add field which is dependent on the depenent field (if present)
       if (dependentFieldValue?.value) {
+        //Add dependent field with selected value
+        this.addDependentField(dependentField, parentField.value, dependentFieldValue?.value);
+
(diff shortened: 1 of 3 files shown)


src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -2692,20 +2692,19 @@ export class AddEditExpensePage implements OnInit {
     this.fg.controls.project.valueChanges
       .pipe(
         takeUntil(this.onPageExit$),
-        filter((project) => !!project),
```
yksolanki9, line 2695: project value will be `null` when uses selects `None` in the UI, so we'll have to clear the existing dependent fields

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -4365,16 +4364,22 @@ export class AddEditExpensePage implements OnInit {
       const dependentFieldValue = txCustomProperties.find(
         (customProp) => customProp.name === dependentField.field_name
       );
-      //Add dependent field with selected value
-      this.addDependentField(dependentField, parentField.value, dependentFieldValue?.value);
 
-      //Add field which is dependent on the depenent field (if present)
       if (dependentFieldValue?.value) {
+        //Add dependent field with selected value
+        this.addDependentField(dependentField, parentField.value, dependentFieldValue?.value);
```
yksolanki9, line 4370: Add the dependent field directly only if it has a value, else check if there are valid values for this field based on the value of parent field

Julias0 (approved):

PR #1732, opened 2023-02-28
Code Coverage : Personal Card Service #1

jayfyle:
# Edit 1: Coverage at 27

personal_card_27


src/app/core/services/personal-cards.service.spec.ts
```diff
@@ -1,16 +1,176 @@
...
+
+    personalCardsService.deleteAccount(accountId).subscribe((res) => {
+      expect(res).toEqual(deletePersonalCardRes);
+      expect(expenseAggregationService.delete).toHaveBeenCalledOnceWith(`/bank_accounts/${accountId}`);
+      done();
+    });
+  });
+
+  it('convertFilters(): should convert selected filters', () => {
+    expect(personalCardsService.convertFilters(selectedFilters1)).toEqual({
```
jayfyle, line 140: Will create a separate file for this output later, the model does not exist, it has been declared as partial type in the service itself. I will move that in a file as well and reorganise once more data has accumulated.

Dimple16: add a `TODO`

Dimple16 (approved):

PR #1739, opened 2023-03-01
Code Coverage: Personal Card Service #3

jayfyle:
# Edit 1: Coverage at 98%

pcard_98

### DIFF
--- a/src/app/core/services/personal-cards.service.spec.ts
+++ b/src/app/core/services/personal-cards.service.spec.ts
@@ -4,12 +4,14 @@ import { ApiV2Service } from './api-v2.service';
 import { ApiService } from './api.service';
 import { ExpenseAggregationService } from './expense-aggregation.service';
 import { DateService } from './date.service';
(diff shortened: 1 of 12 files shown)


src/app/core/services/personal-cards.service.spec.ts
```diff
@@ -404,4 +408,536 @@ describe('PersonalCardsService', () => {
...
+      const filters = {
+        createdOn: {
+          name: DateFilters.thisWeek,
+        },
+      };
+
+      const filterPills = [];
+
+      //@ts-ignore
+      personalCardsService.generateDateFilterPills('createdOn', filters, filterPills);
```
Dimple16, line 423: why are we calling the method in this way?

Dimple16 (approved):
one question

src/app/core/services/personal-cards.service.spec.ts
jayfyle: @Dimple16. This method does not return any value, it modifies the `filterPills` array that is passed to it. Hence I am only calling the function and checking `filterPills` again after method call to make sure that the function is working correctly.

PR #1740, opened 2023-03-01
Fyle 85zrnrbyr Code Coverage 100% : Invitation Requests Service

CODEX108:

### DIFF
--- a/src/app/core/services/invitation-requests.service.spec.ts
+++ b/src/app/core/services/invitation-requests.service.spec.ts
@@ -21,7 +21,7 @@ describe('InvitationRequestsService', () => {
     const expectedEndpoint = '/invitation_requests/invite';
     const expectedRequestBody = { email: mockEmail };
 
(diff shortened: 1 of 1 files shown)


src/app/core/services/invitation-requests.service.spec.ts
```diff
@@ -1,16 +1,30 @@
...
   });
 
-  it('should be created', () => {
-    expect(service).toBeTruthy();
+  it('upsertRouter(): should call the post method with the correct endpoint and email', () => {
+    const mockEmail = 'manjiri.c@fyle.in';
+    const expectedEndpoint = '/invitation_requests/invite';
+    const expectedRequestBody = { email: mockEmail };
+
+    routerApiServiceSpy.post.and.returnValue(of({}));
```
jayfyle, line 24: Check if post actually returns empty object or null, most post calls return `null`.

jayfyle (approved):
Rest looks good.

Dimple16 (approved):

PR #1761, opened 2023-03-09
Code Coverage: Split Expense Service #3

jayfyle:
# Edit 1: Coverage at 80%

split_expense_80


src/app/core/services/split-expense.service.spec.ts
```diff
@@ -236,6 +263,18 @@ describe('SplitExpenseService', () => {
     });
   });
 
+  it('createSplitTxns(): should create split transaction when IDs are not present', (done) => {
+    spyOn(splitExpenseService, 'createTxns').and.returnValue(of(splitTxn2));
+
+    const amount = 16428.56;
+
+    splitExpenseService.createSplitTxns(createSourceTxn2, amount, splitTxn2).subscribe((res) => {
+      expect(res).toEqual(expectedSplitTxns);
```
jayfyle, line 272: Here I have not added `toHaveBeenCalled`, because it fails on the same being used for `createTxn` method. Please ignore the test for now. Will fix it when working on the failing spec method.

src/app/core/services/split-expense.service.ts
```diff
@@ -207,7 +207,7 @@ export class SplitExpenseService {
     );
   }
 
-  runPolicyCheck(etxns: Expense[], fileObjs: FileObject[]): Observable<PolicyViolationTxn> {
+  runPolicyCheck(etxns: Expense[], fileObjs?: FileObject[]): Observable<PolicyViolationTxn> {
```
Dimple16, line 210: `fileObjs` will always be present no? either an empty list or list of files?

Dimple16: we are setting it's default value in `split-expenses.page.ts` - `fileObjs: any[];`

Dimple16 (changes requested):

PR #1763, opened 2023-03-09
test: Code coverage: Merge expenses service - 50%

Dimple16:
click up - https://app.clickup.com/t/85zrpdg2b

Screenshot 2023-03-14 at 1 46 27 AM


src/app/core/models/v1/extended-project.model.ts
```diff
@@ -1,7 +1,7 @@
 export interface ProjectV1 {
   id: number;
-  created_at: Date;
-  updated_at: Date;
+  created_at: Date | string; // API returns a string, but we convert it to a Date
+  updated_at: Date | string; // API returns a string, but we convert it to a Date
```
rvab, line 4: Shouldn't we just use `Date`? Why add string?

src/app/fyle/merge-expense/merge-expense.page.ts
```diff
@@ -628,7 +628,7 @@ export class MergeExpensePage implements OnInit {
   }
 
   onPaymentModeChanged() {
-    this.mergeExpensesService.getCardCardTransactions(this.expenses).subscribe((txns) => {
+    this.mergeExpensesService.getCorporateCardTransactions(this.expenses).subscribe((txns) => {
```
rvab, line 631: `done` is missing

rvab (approved):
minor comments

src/app/core/models/v1/extended-project.model.ts
Dimple16: this is because of the way we use get projects method, API returns a string, but, we convert it to a Date, since for both these cases the same model is used, added `Date | string`

src/app/fyle/merge-expense/merge-expense.page.ts
Dimple16: I have renamed the method name in the component, this isn't a test file

PR #1764, opened 2023-03-09
test: Code Coverage: Auth Service #1

jayfyle:
### Clickup: https://app.clickup.com/t/85zrq04gd

# Edit 1: Coverage at 86

auth_86


src/app/core/services/auth.service.spec.ts
```diff
@@ -1,16 +1,148 @@
...
+    };
+
+    apiService.post.withArgs('/auth/logout', payload).and.returnValue(of(true));
+    apiService.post.withArgs('/auth/logout').and.returnValue(of(true));
+
+    authService
+      .logout(payload)
+      .pipe(
+        finalize(async () => {
+          expect(storageService.delete).toHaveBeenCalledOnceWith('recentlyUsedProjects');
```
jayfyle, line 140: @Dimple16 , I have a doubt here, I was trying to cover the `finalize` block, and I was able to get the coverage but here there should be many calls to storage service with different parameters but they do not happen. The test fails stating this is the only call that went through

src/app/core/services/auth.service.spec.ts
```diff
@@ -1,16 +1,148 @@
...
+      expect(res).toEqual(eouRes3);
+      expect(apiService.get).toHaveBeenCalledOnceWith('/eous/current');
+      expect(dataTransformService.unflatten).toHaveBeenCalledOnceWith(eouFlattended);
+      expect(storageService.set).toHaveBeenCalledOnceWith('user', eouRes3);
+      done();
+    });
+  });
+
+  describe('getRoles():', () => {
+    it('should get roles from access token', (done) => {
```
Dimple16, line 95: mocking the try-catch block is missing, you will have to spyOn JSON, 'parse'

src/app/core/services/auth.service.spec.ts
Dimple16: try this method:
```
it('should logout and delete all storage items', fakeAsync(() => {
spyOn(storageService, 'delete').and.returnValue(Promise.resolve());
authService
      .logout(payload).subscribe(() => {
       expect(apiService.post).toHaveBeenCalledWith('/auth/logout', Payload);
       expect(storageService.delete).toHaveBeenCalledWith('recentlyUsedProjects');
       expect(storageService.delete).toHaveBeenCalledWith('recentlyUsedCategories');
       ....and so on...
       tick();
  }));
```

Dimple16 (changes requested):
added comments

src/app/core/services/auth.service.spec.ts
Dimple16: this isn't taken care of yet?

src/app/core/services/auth.service.spec.ts
```diff
@@ -1,16 +1,158 @@
...
+      finalize(() => {
+        expect(storageService.delete).toHaveBeenCalledOnceWith('recentlyUsedProjects');
+        expect(storageService.delete).toHaveBeenCalledOnceWith('recentlyUsedCategories');
+        expect(storageService.delete).toHaveBeenCalledOnceWith('recentlyUsedMileageCategories');
+        expect(storageService.delete).toHaveBeenCalledOnceWith('recentlyUsedPerDiemCategories');
+        expect(storageService.delete).toHaveBeenCalledOnceWith('recentlyUsedCostCenters');
+        expect(storageService.delete).toHaveBeenCalledOnceWith('user');
+        expect(storageService.delete).toHaveBeenCalledOnceWith('role');
+        expect(storageService.delete).toHaveBeenCalledOnceWith('currentView');
+        expect(storageService.delete).toHaveBeenCalledOnceWith('ui-grid-pagination-page-size');
```
Dimple16, line 147: we don't use these, you should remove these from here,  `ui-grid-pagination-page-size`, 
`ui-grid-pagination-page-number`, `customExportFields` etc, are never used,
can you individually check what are used and keep them? remove the unused code

Dimple16 (changes requested):
added comments

src/app/core/services/auth.service.spec.ts
```diff
@@ -1,16 +1,173 @@
...
+      .logout(payload)
+      .pipe(
+        tap((res) => {
+          expect(res).toBeTruthy();
+          expect(apiService.post).toHaveBeenCalledWith('/auth/logout');
+          expect(apiService.post).toHaveBeenCalledWith('/auth/logout', payload);
+          expect(apiService.post).toHaveBeenCalledTimes(2);
+        }),
+        finalize(() => {
+          expect(storageService.delete).toHaveBeenCalledWith('recentlyUsedProjects');
```
Dimple16, line 167: add a `Todo` to revisit this section

Dimple16 (approved):

PR #1779, opened 2023-03-13
fix: Dependent Fields: Fix issues found during testing

yksolanki9:
Clickup link -> https://app.clickup.com/t/85zrrfb7t

Fixes:
- Check symbol will be shown only for the last selected item
- Show dependent fields value inside modal in ascending order
- Always show None option in the modal
- Selected value of the dependent field will always be shown irrespective if it appears in search
- Fix size and top padding of validation message
- Don't trigger valueChanges if user selects the same value

Optimizations:
- Avoid multiple api calls by caching last 5 values of dependent fields, so that if user comes back to parent field after selecting child field, we don't make one more api call as the response is already cached.
- Passed empty string as search query in expense forms to avoid duplicate calls

UI Preview:
Screenshot 2023-03-13 at 7 54 58 PM


src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -4278,12 +4279,13 @@ export class AddEditExpensePage implements OnInit {
...
           (dependentCustomField) => dependentCustomField.parent_field_id === parentFieldId
         );
-        if (dependentField) {
+        if (dependentField && parentFieldValue) {
           return this.dependentFieldsService
             .getOptionsForDependentField({
               fieldId: dependentField.id,
               parentFieldId,
               parentFieldValue,
+              searchQuery: '',
```
yksolanki9, line 4288: There are two instances where this api call is made.
1. When user changes parent field
2. When user open modal for the child field

In case 2, we're passing searchquery as empty string using `startsWith` operator to get all dependent fields which is missing in 1, and hence the returned value is not from the  cached response, but another api call is made. 

Passing empty string here ensures a cache hit

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -4304,9 +4306,11 @@ export class AddEditExpensePage implements OnInit {
       value: [value, (dependentField.is_mandatory || null) && Validators.required],
     });
 
-    dependentFieldControl.valueChanges.pipe(takeUntil(this.onPageExit$)).subscribe((value) => {
-      this.onDependentFieldChanged(value);
-    });
+    dependentFieldControl.valueChanges
+      .pipe(takeUntil(this.onPageExit$), distinctUntilKeyChanged('value'))
```
yksolanki9, line 4310: - Here, `distinctUntilKeyChanged()` works only when user has selected the value for the field, and it doesn't work in edit expense case when the user updates the field for the first time as we use `emitEvent: false` during patchValue.
- I also tried using custom comparator here, but it didn't work as well

Let me know if you have any ideas on how to fix this

Julias0: For `edit` case, can we just trigger `this.onDependentFieldChanged(value)` manually first?

PR #1782, opened 2023-03-14
test: Code coverage - Token service - 100% branches

Dimple16:
click up - https://app.clickup.com/t/85zrpu3mx

Screenshot 2023-03-14 at 1 03 12 PM

### DIFF
--- a/src/app/core/services/token.service.spec.ts
+++ b/src/app/core/services/token.service.spec.ts
@@ -1,4 +1,4 @@
-import { TestBed } from '@angular/core/testing';
+import { fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { callbackify } from 'util';
 import { SecureStorageService } from './secure-storage.service';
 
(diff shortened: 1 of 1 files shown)


src/app/core/services/token.service.spec.ts
```diff
@@ -1,16 +1,107 @@
...
-    expect(service).toBeTruthy();
+    expect(tokenService).toBeTruthy();
+  });
+
+  it('getAccessToken(): should get access token from secure storage', (done) => {
+    secureStorageService.get.and.returnValue(Promise.resolve('token'));
+    tokenService.getAccessToken().then((token) => {
+      expect(token).toEqual('token');
+      expect(secureStorageService.get).toHaveBeenCalledOnceWith('X-AUTH-TOKEN');
+      done();
```
rvab, line 37: iirc, we should use `tick` for promises. Could you check this once?

src/app/core/services/token.service.spec.ts
```diff
@@ -1,16 +1,107 @@
...
+    });
+  });
+
+  it('setAccessToken(): should set access token to secure storage', (done) => {
+    userEventService.setToken.and.returnValue();
```
rvab, line 42: is this intentional? returning empty value?

rvab (changes requested):
One important comment

src/app/core/services/token.service.spec.ts
Dimple16: fixed

src/app/core/services/token.service.spec.ts
Dimple16: missed `null`, fixed

rvab (approved):

PR #1802, opened 2023-03-17
test: Code coverage - Utility service - 64% statements

Dimple16:
click up - https://app.clickup.com/t/85zrpv5pr

Screenshot 2023-03-17 at 9 01 41 PM

Note: Skipping tests for methods involving callbacks in this PR


src/app/core/services/utility.service.spec.ts
```diff
@@ -1,16 +1,131 @@
...
+  });
+
+  it('should discard null characters', () => {
+    const mockStr = 'Fyle\u0000 Expense!\u0000';
+    expect(utilityService.discardNullChar(mockStr)).toEqual(
+      mockStr.replace(/[\u0000][\u0008-\u0009][\u000A-\u000C][\u005C]/g, '')
+    );
+  });
+
+  describe('refineNestedObject():', () => {
```
rvab, line 33: didn't get what is the use of this method. Is the output the same as the input?

rvab (approved):

src/app/core/services/utility.service.spec.ts
Dimple16: it specifically removes/discards null characters for custom fields of type `SELECT`

PR #1824, opened 2023-03-23
fix: Code Coverage: Launch Darkly Service #2

jayfyle:
### Clickup : https://app.clickup.com/t/85zrt3mxb

### DIFF
--- a/src/app/core/services/launch-darkly.service.spec.ts
+++ b/src/app/core/services/launch-darkly.service.spec.ts
@@ -77,7 +77,7 @@ describe('LaunchDarklyService', () => {
     });
   });
 
-  it('shutDownClient(): should shut down client', () => {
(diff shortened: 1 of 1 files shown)


src/app/core/services/launch-darkly.service.spec.ts
```diff
@@ -43,13 +43,13 @@ describe('LaunchDarklyService', () => {
     expect(launchDarklyService.isTheSameUser(lDUser)).toBeFalse();
   });
 
-  it('updateCache(): should update cache', () => {
+  xit('updateCache(): should update cache', () => {
```
Dimple16, line 46: instead, remove these tests, we'll never enable them again anyway

Dimple16 (changes requested):

Dimple16 (approved):

PR #1843, opened 2023-03-28
fix: Sign In Page Google Button Loader

jayfyle:
# Sign in with Google Button Loader fixed

### DIFF
--- a/src/app/auth/sign-in/sign-in.page.scss
+++ b/src/app/auth/sign-in/sign-in.page.scss
@@ -83,8 +83,7 @@ $password-icon: #b9beba;
     }
     margin-top: 24px;
     &-btn {
-      display: flex;
-      flex-wrap: wrap;
+      display: inline-block;
       align-items: center;
       justify-content: center;
       gap: 10px;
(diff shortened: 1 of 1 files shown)


yksolanki9 (changes requested):
https://caniuse.com/css-display-contents
`display: contents` is not fully supported by all browsers. Can we try a different approach?

Julias0 (changes requested):
Will review after Yash Solankis approval

yksolanki9 (approved):

src/app/auth/sign-in/sign-in.page.scss
```diff
@@ -83,7 +83,7 @@ $password-icon: #b9beba;
     }
     margin-top: 24px;
     &-btn {
-      display: flex;
+      display: inline-block;
       align-items: center;
```
yksolanki9, line 87: are align-items, justify-content and gap needed?

Julias0 (approved):

yksolanki9 (approved):

PR #1833, opened 2023-03-24
feat: Optimization for setupFormInit

suyashpatil78:
method added to category service for fetching id specific category

Clickup Link - https://app.clickup.com/t/85zrtz58p

PR #1869, opened 2023-03-31
feat: Optimizations for checkPolicyViolation

suyashpatil78:
Clickup - https://app.clickup.com/t/85zruru5j


suyashpatil78:
Coverage
Screenshot 2023-03-31 at 2 55 26 PM

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -2822,11 +2822,11 @@ export class AddEditExpensePage implements OnInit {
...
     if (!transactionCopy.org_category_id) {
-      return this.categoriesService.getAll().pipe(
-        map((categories: OrgCategory[]) => {
-          const unspecifiedCategory = categories.find(
-            (category) => category?.fyle_category?.toLowerCase() === 'unspecified'
-          );
+      const categoryName = 'Unspecified';
+      return this.categoriesService.getCategoryByName(categoryName).pipe(
+        map((category: OrgCategory) => {
+          console.log(category);
```
yksolanki9, line 2828: Remove console log

src/app/core/services/categories.service.ts
```diff
@@ -35,6 +35,20 @@ export class CategoriesService {
...
+  getCategoryByName(name: string): Observable<OrgCategory> {
+    const data = {
+      params: {
+        name: 'eq.' + name,
+      },
+    };
+    return this.spenderPlatformV1ApiService.get<PlatformApiResponse<PlatformCategory>>(`/categories`, data).pipe(
+      map((res) => this.transformFrom(res.data)),
+      map((res) => this.addDisplayName(res)),
+      map((responses) => responses.find((response) => response.name === name))
```
yksolanki9, line 48: why do we need to find by name again?

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
suyashpatil78: @yksolanki9 done

src/app/core/services/categories.service.ts
```diff
@@ -35,6 +35,20 @@ export class CategoriesService {
     );
   }
 
+  @Cacheable()
+  getCategoryByName(name: string): Observable<OrgCategory> {
+    const data = {
+      params: {
+        name: 'eq.' + name,
```
yksolanki9, line 42: Confirm with backend team if the first name should be in upper or lower case?

yksolanki9: Will it be better to use `ilike` to make it case insensitive?

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -2822,11 +2822,10 @@ export class AddEditExpensePage implements OnInit {
...
     if (!transactionCopy.org_category_id) {
-      return this.categoriesService.getAll().pipe(
-        map((categories: OrgCategory[]) => {
-          const unspecifiedCategory = categories.find(
-            (category) => category?.fyle_category?.toLowerCase() === 'unspecified'
-          );
+      const categoryName = 'Unspecified';
+      return this.categoriesService.getCategoryByName(categoryName).pipe(
+        map((category: OrgCategory) => {
+          const unspecifiedCategory = category;
```
yksolanki9, line 2828: Another variable is not needed here

yksolanki9 (changes requested):

src/app/core/services/categories.service.ts
suyashpatil78: @yksolanki9 The data is coming in the form of array, therefore we need to get single response from it.
We have done same thing in getCategoryById - https://github.com/fylein/fyle-mobile-app/pull/1833/files#diff-00ea7bf26983400bc3a2227bb4c1695132c71429b10ebea74274d01ddf41e6eaR180

src/app/core/services/categories.service.ts
suyashpatil78: The data could be in uppercase and lowercase, both. So it is better to use `ilike`

src/app/core/services/categories.service.ts
```diff
@@ -35,6 +35,20 @@ export class CategoriesService {
     );
   }
 
+  @Cacheable()
+  getCategoryByName(name: string): Observable<OrgCategory> {
+    const data = {
+      params: {
+        name: 'ilike.%' + name + '%',
```
Julias0, line 42: if we are searching by the whole name, lets avoid using `%` for wild card characters, since in certain cases it might lead to wrong results.

Example - 
You have `Lunch` and `Team Lunch`. If you search with the above, you run the risk of getting team lunch when searching for lunch

suyashpatil78: Got it!

yksolanki9: Yeah I meant ilike without the matching operator

src/app/core/services/categories.service.ts
yksolanki9: Do you need to find an element inside an array which has only one element?

yksolanki9: Yes, this was what I was hinting towards

yksolanki9 (approved):

Julias0 (approved):

PR #1888, opened 2023-04-03
test: Code Coverage: Reports Card Component

jayfyle:
### Clickup: https://app.clickup.com/t/85zrv3w9b

# Coverage at 100%.
MyReportsCard and TeamReportsCard component removed and replaced with ReportCard component in shared directory


capacitor.config.json
```diff
@@ -0,0 +1 @@
+{}
```
Dimple16, line 1: why is this here?

Dimple16 (approved):
minor comment, fix and merge

PR #1926, opened 2023-04-13
fix: removed push notification plugins and related service

SowmyaFyle:
app.clickup.com


yksolanki9 (approved):

Julias0 (changes requested):
Sending back since this wont be merged

PR #1931, opened 2023-04-14
test: Code Coverage: Route Selector Component

jayfyle:
### Clickup: https://app.clickup.com/t/85zrw6gtd

# Coverage at 97

route_selector_97


src/app/shared/components/route-selector/route-selector.component.spec.ts
```diff
@@ -1,26 +1,238 @@
...
-    })
-  );
+  beforeEach(waitForAsync(() => {
+    const modalControllerSpy = jasmine.createSpyObj('ModalController', ['create']);
+    const injectorSpy = jasmine.createSpyObj('Injector', ['get']);
+    TestBed.configureTestingModule({
+      declarations: [RouteSelectorComponent],
+      imports: [IonicModule.forRoot(), MatCheckboxModule, ReactiveFormsModule, MatIconTestingModule, MatIconModule],
+      providers: [
+        NgControl,
```
Julias0, line 30: Is ngcontrol required since we removed off the dependency

src/app/shared/components/route-selector/route-selector.component.spec.ts
```diff
@@ -1,26 +1,238 @@
...
 
   it('should create', () => {
     expect(component).toBeTruthy();
   });
+
+  it('customDistanceValidator(): should validate distance form control', () => {
+    component.form.controls.distance.setValue(10);
+    fixture.detectChanges();
+    const result = component.customDistanceValidator(component.form.controls.distance);
+    expect(result).toBeNull();
```
Julias0, line 69: shouldnt we check for the error case too?

src/app/shared/components/route-selector/route-selector.component.spec.ts
```diff
@@ -1,26 +1,238 @@
...
+
+  it('registerOnChange(): should register changed value', async () => {
+    const changeTestCallback = jasmine.createSpyObj('changeTested', ['test']);
+    component.registerOnChange(changeTestCallback.test);
+
+    component.form.controls.distance.setValue(10);
+
+    fixture.whenStable();
+
+    expect(changeTestCallback.test).toHaveBeenCalled();
```
Julias0, line 98: We should also check the value with which this has been called

src/app/shared/components/route-selector/route-selector.component.spec.ts
```diff
@@ -1,26 +1,238 @@
...
+
+  it('registerOnTouched(): should registered onTouched property', () => {
+    const touchFn = () => {
+      component.form.markAsDirty();
+    };
+    component.registerOnTouched(touchFn);
+
+    fixture.whenStable();
+
+    expect(component.onTouched).toEqual(touchFn);
```
Julias0, line 109: We shouldnt compare objects. Use the DOM to trigger clicked event on the input element and check if the callback inside `registerOnTouched` was called or not

src/app/shared/components/route-selector/route-selector.component.spec.ts
```diff
@@ -1,26 +1,248 @@
...
+  it('registerOnTouched(): should registered onTouched property', async () => {
+    const onTouchTested = jasmine.createSpyObj('touchedTested', ['touchedTested']);
+
+    component.registerOnTouched(onTouchTested.touchTested);
+
+    expect(onTouchTested.touchedTested).not.toHaveBeenCalled();
+
+    const input = getElementBySelector(fixture, '.route-selector--input');
+    input.dispatchEvent(new Event('blur'));
+
```
jayfyle, line 118: Over here the matcher for `onTouchTested.touchedTested` does not work. Somehow the blur event does not trigger the function, so the spec fails as the function never gets called. All other comments resolved.

Julias0 (approved):

PR #1944, opened 2023-04-20
test: Code coverage BankAccountCards Component

CODEX108:
clickUp: https://app.clickup.com/t/85zrxn9qg

image


src/app/shared/components/bank-account-cards/bank-account-cards.component.spec.ts
```diff
@@ -1,24 +1,61 @@
...
+      fixture.detectChanges();
+      expect(component.changed.emit).toHaveBeenCalledOnceWith('baccBlpSkgBbN0');
+    });
+
+    it('should emit undefined if the selected card is not found in linkedAccounts', () => {
+      spyOn(component.changed, 'emit');
+      component.linkedAccounts = [deletePersonalCardRes];
+      component.onCardChange([{ realIndex: 1 }]);
+      fixture.detectChanges();
+      expect(component.changed.emit).toHaveBeenCalledOnceWith(undefined);
```
Julias0, line 50: Cant this ever happen in a real scenario?

Julias0 (changes requested):
Added 1 comment

src/app/shared/components/bank-account-cards/bank-account-cards.component.spec.ts
CODEX108: This won't happen in a real scenario because in any case the `id` is always present for card data and then it won't call undefined because the linkedAcc array ideally contains the cards (which are selected) in the above test case since I have provided the data of a deleted card, the emit method was called with `undefined`, I wanted to cover the  `?.` opr, therefore, I had added this test case.

Julias0 (approved):

PR #1949, opened 2023-04-21
feat: Update github actions to use the latest build stack

yksolanki9:
Clickup - https://app.clickup.com/t/85zrxy1zr


.github/workflows/appflow.yml
yksolanki9, line 1: This action was not used anywhere so deleted it

Julias0 (approved):

PR #1983, opened 2023-04-28
test: Tests for dependent fields component

yksolanki9:
Clickup - https://app.clickup.com/t/85zry9tj2

Code coverage:
Screenshot 2023-05-03 at 5 35 56 PM


src/app/shared/components/dependent-fields/dependent-fields.component.spec.ts
```diff
@@ -1,24 +1,517 @@
...
+      component.onDependentFieldChanged(data);
+      fixture.detectChanges();
+
+      expect(component.removeAllDependentFields).toHaveBeenCalledOnceWith(0);
+      expect(component.getDependentField).toHaveBeenCalledOnceWith(data.id, data.value);
+
+      component.getDependentField(data.id, data.value).subscribe((result) => {
+        expect(component.isDependentFieldLoading).toBeFalse();
+        expect(result).toEqual(null);
+        expect(component.addDependentField).not.toHaveBeenCalled();
```
Julias0, line 513: subscribes neeed to have the done callback called inside them or we need to use fakeAsync otherwise the subscribe block is not executed and the assertions not triggered

Julias0: Same in the rest of the cases

src/app/shared/components/dependent-fields/dependent-fields.component.spec.ts
```diff
@@ -1,24 +1,517 @@
...
+    component.onPageExit$ = new Subject();
   }));
 
   it('should create', () => {
     expect(component).toBeTruthy();
   });
+
+  it('ngOnDestroy(): should not call complete if onPageExit$ is not defined', () => {
+    component.onPageExit$ = null;
+    const onPageExitSpy = jasmine.createSpyObj('onPageExit$', ['next', 'complete']);
```
Julias0, line 69: use SpyOn to do this instead of creating a spy here and not passing it in. SpyOn component.onPageExit$ and then set it to null. Call component.ngOnDestroy() and check if the next and complete got called

src/app/shared/components/dependent-fields/dependent-fields.component.spec.ts
```diff
@@ -1,24 +1,517 @@
...
+        },
+      };
+      component.onPageExit$ = null;
+
+      spyOn(component, 'addDependentFieldWithValue').and.stub();
+    });
+
+    it('should initialize onPageExit$ and add dependent field when it is the first change', () => {
+      component.ngOnChanges(changes);
+      expect(component.onPageExit$).toEqual(new Subject());
```
Julias0, line 103: Didnt you set this to null in the beforeEach? Why compare this to new Subject here?

Julias0 (changes requested):
Added few comments

src/app/shared/components/dependent-fields/dependent-fields.component.spec.ts
yksolanki9: Still getting this error - ` TypeError: Cannot read properties of null (reading 'next')` 
We can discuss this once you're back

yksolanki9: This is resolved now, I used the return value of spy instead

Julias0 (approved):

PR #1991, opened 2023-05-04
feat: Camera Service

jayfyle:
## Changes made:
1. Removed all imports for `@capacitor/camera` from components CaptureReceipt, CameraPreview
2.  Added the function `requestCameraPermission` which the only function used from the plugin


src/app/core/services/camera.service.ts
```diff
@@ -0,0 +1,13 @@
+import { Injectable } from '@angular/core';
+import { Camera, CameraPermissionType } from '@capacitor/camera';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class CameraService {
+  constructor() {}
+
+  requestCameraPermissions(permissionsType?: CameraPermissionType[]) {
```
Julias0, line 10: we can remove the nullable here

src/app/core/services/camera.service.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { TestBed } from '@angular/core/testing';
+
+import { CameraService } from './camera.service';
+
+describe('CameraService', () => {
```
Julias0, line 5: can we add unit tests for this

Julias0 (changes requested):
Added comments

src/app/core/services/camera.service.ts
jayfyle: It has to be optional as it defined as an optional in the plugin itself.  We are using this method twice, once with params and once without them.

Julias0 (approved):

PR #2014, opened 2023-05-12
fix: Fixed split_group_id incorrectly set for the first split in creation flow #3806

OmkarJ13:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=jgL6QnL)

ClickUp: https://app.clickup.com/t/85zt1cxkc

Loom Video: 
https://www.loom.com/share/8dee01c9e9f0464cbe4da5f0f286d386

Description: 
- When we split an expense from the add expense screen itself, since there is no original expense that these three expenses originate from, the first split is created first with the `split_group_id` equal to its `tx_id`
- And the rest of the splits are created later, these have the `split_group_id` equal to the first split’s `tx_id`.
- This leads to the first split not being recognized as a split expense, because `split_group_id` == `tx_id` and for split we need this to be not equal
- So in this case, we are generating a new expense id and setting it as the `split_group_id` to all the expenses that were created from add expense flow, to make sure all of them are identified as a split.
- Full discussion thread: https://fylein.slack.com/archives/C03QJNCAFD5/p1683700355251629?thread_ts=1683700033.982049&cid=C03QJNCAFD5


src/app/core/mock-data/transaction.data.ts
```diff
@@ -2277,12 +2277,12 @@ export const expectedSplitTxns: Transaction[] = [
     user_can_delete: true,
     txn_dt: new Date('2023-03-02T18:30:00.000Z'),
     category: null,
-    amount: 80,
+    amount: 120,
```
OmkarJ13, line 2280: fixed expected split transactions, before both of them were the same, before both of them were the same and this was working in previous code because we were manipulating the underlying `splitExpenses` array with splice, since we don't use it anymore, fixed this

src/app/core/mock-data/transaction.data.ts
```diff
@@ -2264,378 +2264,3 @@ export const txnParam2: Transaction = {
   categoryDisplayName: 'Food / Travelling - Inland',
   custom_attributes: null,
 };
-
-export const expectedSplitTxns: Transaction[] = [
```
OmkarJ13, line 2268: The expected split expenses data was incorrect due to the way the response was mocked and the usage of `splice` in earlier code which was mutating the mocked response object.

I have fixed this and we directly check if the expected response equals the mocked response, like the other methods in this service.

So, this isn't needed.

OmkarJ13 (commented):
Some test with TeamAdvanceCard is failing, this looks existing on master
Edit: This is fixed on master, pulled master to this branch.

src/app/core/services/utility.service.spec.ts
```diff
@@ -52,6 +52,35 @@ describe('UtilityService', () => {
     });
   });
 
+  describe('generateRandomString():', () => {
+    it('should return a string of the specified length', () => {
+      const result = utilityService.generateRandomString(10);
```
abhishek1234321, line 57: should we check if the returned type is string?

src/app/core/services/utility.service.spec.ts
```diff
@@ -52,6 +52,35 @@ describe('UtilityService', () => {
...
+
+    it('should only contain valid characters', () => {
+      const result = utilityService.generateRandomString(10);
+      const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+      for (let i = 0; i < result.length; i++) {
+        expect(characters).toContain(result[i]);
+      }
+    });
+
+    it('should return an empty string when length is 0 or negative', () => {
```
abhishek1234321, line 75: these should be 2 different test cases

abhishek1234321 (approved):

src/app/core/services/utility.service.spec.ts
OmkarJ13: Fixed 👍

rvab (approved):

src/app/core/services/utility.service.spec.ts
OmkarJ13: I think it doesn't make a lot of sense, since in typescript we restrict the function to only return `string`.

The only case I can think of is if we were returning `null` or `undefined` in the failure case, but since here we return an empty string I can't think of any value added by doing this.

PR #2067, opened 2023-05-31
fix: Back Button Service

jayfyle:
### Clickup: https://app.clickup.com/t/85zt699cp

# Fixed types

### DIFF
--- a/src/app/core/services/back-button.service.ts
+++ b/src/app/core/services/back-button.service.ts
@@ -33,11 +33,17 @@ export class BackButtonService {
       .pipe(
         tap((exitAppPopover) => exitAppPopover.present()),
         switchMap((exitAppPopover) => exitAppPopover.onWillDismiss()),
-        map((popoverDetails) => {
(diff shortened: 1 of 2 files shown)


src/app/core/services/back-button.service.ts
```diff
@@ -33,11 +33,17 @@ export class BackButtonService {
...
-        map((popoverDetails) => {
-          if (popoverDetails?.data?.action === 'close') {
-            return App.exitApp();
+        map(
+          (popoverDetails: {
+            data?: {
+              action?: string;
+            };
+          }) => {
+            if (popoverDetails?.data?.action === 'close') {
```
yksolanki9, line 42: Every modal/popover in ionic returns data object. We can create a generic type for this.
So, the type here would look like `OverlayResponse<{action?: string}>`
Let me know if you need any help with this

yksolanki9 (changes requested):

Dimple16 (approved):

yksolanki9 (approved):

PR #2198, opened 2023-07-18
fix: remove unused images

vaibhav-fyle:
Clickup [task](https://app.clickup.com/t/863h7ma3b)
Slack [thread](https://fylein.slack.com/archives/C05E63Q5MQ8/p1688711032892449)


yksolanki9 (approved):
Check if splashscreen and app drawer icon is visible on both android and ios before merging

PR #2224, opened 2023-08-01
feat: Upgrade Mobile App to Latest Capacitor

rupam-seal:
### Description
copilot:summary

copilot:poem

### Walkthrough
copilot:walkthrough

## Clickup
https://app.clickup.com/t/85ztp0vrt

## Code Coverage
Please add code coverage here

## UI Preview


package.json
```diff
@@ -39,24 +39,24 @@
     "@awesome-cordova-plugins/in-app-browser": "^6.2.0",
```
OmkarJ13, line 39: Do we not need to upgrade android project variables in files like `variables.gradle` and `build.gradle`? Also do we not need to update the plugins?

src/app/core/services/device.service.ts
```diff
@@ -21,10 +21,11 @@ export class DeviceService {
     }).pipe(
       map(({ deviceInfo, deviceId, appInfo }) =>
         Object.assign(deviceInfo, deviceId, {
+          uuid: '',
```
OmkarJ13, line 24: Why are we setting this to an empty string?

OmkarJ13 (changes requested):
Can you check the breaking changes mentioned here and make changes as suggested by capacitor
https://capacitorjs.com/docs/updating/5-0

Dimple16 (changes requested):
I'll review once the lint and unit tests are resolved, please fix and reassign

Julias0 (changes requested):
Will review after Dimple

PR #2628, opened 2023-12-01
fix: Billable field not auto-filling the default value

OmkarJ13:
### Description
copilot:summary

copilot:poem

### Walkthrough
copilot:walkthrough

## Clickup
https://app.clickup.com/t/86ctyqe5t

### DIFF
--- a/src/app/core/services/expense-fields.service.ts
+++ b/src/app/core/services/expense-fields.service.ts
@@ -12,6 +12,7 @@ import { OrgCategory } from '../models/v1/org-category.model';
 import { AuthService } from './auth.service';
 import { DateService } from './date.service';
 import { SpenderPlatformV1ApiService } from './spender-platform-v1-api.service';
+import { cloneDeep } from 'lodash';
 
 @Injectable({
   providedIn: 'root',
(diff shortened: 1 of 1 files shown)


src/app/core/services/expense-fields.service.ts
```diff
@@ -210,7 +211,8 @@ export class ExpenseFieldsService {
   }
 
   private formatBillableFields(expenseFields: ExpenseField[]): ExpenseField[] {
-    return expenseFields.map((field) => {
+    const expenseFieldsCopy = cloneDeep(expenseFields);
```
OmkarJ13, line 214: We were storing the expense fields in cache, and object mutation updated the cache, thus whenever retrieving using this method, the default_value was evaluated as false because `true === 'true'` is not true. Only the first time the `'true'/'false' === 'true'` case happened.

src/app/fyle/add-edit-per-diem/add-edit-per-diem.page.ts
```diff
@@ -548,7 +548,14 @@ export class AddEditPerDiemPage implements OnInit {
...
           const control = keyToControlMap[defaultValueColumn];
-          if (!control.value && defaultValueColumn !== 'billable') {
+          if (!control.value && !control.touched && defaultValueColumn !== 'billable') {
+            control.patchValue(defaultValues[defaultValueColumn]);
+          } else if (
+            defaultValueColumn === 'billable' &&
+            this.fg.controls.project.value &&
+            (control.value === undefined || control.value === null) &&
+            !control.touched
+          ) {
```
OmkarJ13, line 558: This check was missing in per diem page, hence it was not getting auto-filled, have added it, have copied this from add-edit expense/mileage page only.

PR #2797, opened 2024-03-01
feat: Added filtering of pending txns in tasks if pending restriction is enabled from org settings.

howdysuraj:
### Description
Added filtering of pending txns in tasks if pending restriction is enabled from org settings.

## Clickup
https://app.clickup.com/t/86cur9q7n

### Test Status
Screenshot 2024-03-04 at 10 21 11 AM

## UI Preview
Screenshot 2024-03-01 at 12 38 37 PM

### DIFF
--- a/src/app/core/services/tasks.service.ts
+++ b/src/app/core/services/tasks.service.ts
@@ -17,6 +17,8 @@ import { TaskDictionary } from '../models/task-dictionary.model';
 import { CorporateCreditCardExpenseService } from './corporate-credit-card-expense.service';
 import { Datum } from '../models/v2/stats-response.model';
 import { ExpensesService } from './platform/v1/spender/expenses.service';
+import { OrgSettingsService } from './org-settings.service';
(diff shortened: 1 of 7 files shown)


src/app/core/services/tasks.service.ts
```diff
@@ -566,18 +568,32 @@ export class TasksService {
...
-    return this.expensesService
-      .getExpenseStats({
-        state: 'in.(COMPLETE)',
-        or: '(policy_amount.is.null,policy_amount.gt.0.0001)',
-        report_id: 'is.null',
+    let queryParams = {
+      state: 'in.(COMPLETE)',
+      or: '(policy_amount.is.null,policy_amount.gt.0.0001)',
+      report_id: 'is.null',
+      and: '()',
```
howdysuraj, line 575: I just added and so that type script not show error when assigning and in line number 586

Julias0 (approved):

PR #2806, opened 2024-03-06
feat: This is the base branch for commute details related changes

suyashpatil78:
### Description
copilot:summary

copilot:poem

### Walkthrough
copilot:walkthrough

## Clickup
app.clickup.com

## Code Coverage
Please add code coverage here

## UI Preview


suyashpatil78:
Would be fixing unit tests separately in subsequent PRs

src/global.scss
```diff
@@ -1002,6 +1002,15 @@ ion-modal.payment-mode-modal {
   }
 }
 
+ion-modal.add-location-modal {
+  &::part(content) {
+    border-radius: 16px 16px 0 0 !important;
```
howdysuraj, line 1007: can we get rid of this !important and try to use the css specificity  for targeting the element if possible

src/app/shared/components/route-selector/route-selector.component.ts
```diff
@@ -85,10 +99,10 @@ export class RouteSelectorComponent implements OnInit, ControlValueAccessor, OnD
     this.onChangeSub.unsubscribe();
   }
 
-  customDistanceValidator(control: AbstractControl) {
+  customDistanceValidator(control: AbstractControl): { invalidDistance: boolean } {
     const passedInDistance = control.value && +control.value;
```
howdysuraj, line 103: @suyashpatil78 are we trying to control.value to number using +, can we use some simple and much readable approach for number conversion , it seems everyone is not aware of this shorthand.

howdysuraj (approved):

src/global.scss
suyashpatil78: actually we don't need !important removing this

src/app/shared/components/route-selector/route-selector.component.ts
suyashpatil78: sure Suraj, we can use `parseFloat`

PR #2898, opened 2024-04-26
feat: add methods to get reports approvers

bistaastha:


src/app/core/services/platform/v1/approver/reports.service.ts
```diff
@@ -12,6 +17,56 @@ export class ApproverReportsService {
...
+      },
+    };
+    return this.approverPlatformApiService.get<PlatformApiResponse<Report>>('/reports', config);
+  }
+
+  getReportsStats(params: PlatformStatsRequestParams): Observable<StatsResponse> {
+    return this.approverPlatformApiService
+      .post<{ data: StatsResponse }>('/reports/stats', {
+        data: {
+          query_params: `state=${params.state}`,
```
suyashpatil78, line 60: Use a variable instead at line 57 and pass it as arguments in the POST call just like you did in `getReportsByParams`

src/app/core/services/platform/v1/approver/reports.service.ts
```diff
@@ -12,6 +17,56 @@ export class ApproverReportsService {
...
+        },
+      })
+      .pipe(map((res) => res.data));
+  }
+
+  getReport(id: string): Observable<Report> {
+    return this.getReportsByParams({ id: `eq.${id}` }).pipe(map((res) => res.data[0]));
```
suyashpatil78, line 67: same as above

suyashpatil78 (approved):
minor comments

PR #2903, opened 2024-04-26
feat: Add new interface PlatformProject + cleanup ExtendedProject unwanted fields

SahilK-027:
### Description
as per the discussion in thread: https://fylein.slack.com/archives/C06M78V7JKD/p1713850203385719
- Remove
```diff
- ap1_email: string;
- ap1_full_name: string;
- ap2_email: string;
- ap2_full_name: string;
- project_approver1_id: string;
- project_approver2_id: string;
```
Removed the above fields from `ExtendedProjects`

- Created `PlatformProject` interface
- Add transform method

## Clickup
[Please add link here](https://app.clickup.com/t/86cv7my4h)


src/app/core/models/v2/extended-project.model.ts
```diff
@@ -1,11 +1,5 @@
 export interface ExtendedProject {
-  ap1_email: string;
-  ap1_full_name: string;
-  ap2_email: string;
-  ap2_full_name: string;
   project_active: boolean;
-  project_approver1_id: string;
-  project_approver2_id: string;
```
devendrafyle, line 8: can we check the usage of this modal, it should not break the previous code.

SahilK-027: We are no longer using these keys.
You can checkout discussion regarding this here.
https://fylein.slack.com/archives/C06M78V7JKD/p1713850203385719

devendrafyle: Instead of this can we make these keys optional or can create a new modal for your use case, removing keys from the existing modal changes the previous model which is not correct imo because the previous modal was created about v2 APIs response that is still correct.

devendrafyle (commented):
Minor comment

src/app/core/models/v2/extended-project.model.ts
SahilK-027: Ok. I will confirm this once.

src/app/core/services/projects.service.ts
```diff
@@ -161,8 +159,43 @@ export class ProjectsService {
...
+      org_id: platformProject.org_id,
+      description: platformProject.description,
+      active: platformProject.is_enabled,
+      org_category_ids: platformProject.category_ids,
+    }));
+
+    return projectV1;
+  }
+
+  transformToV2Response(platformProject: PlatformProject[]): ExtendedProject[] {
```
suyashpatil78, line 184: If this is returns `V2` response, shouldn't the model be named as `ProjectV2[]`. If this requires changing at multiple places, feel free to do it in separate PR.

SahilK-027: So we should replace all the occurrences of  ExtendedProject with ProjectV2, is it?

src/app/core/models/platform/platform-project.model.ts
```diff
@@ -0,0 +1,13 @@
+export interface PlatformProject {
+  id: number;
+  org_id: string;
+  created_at: Date;
+  updated_at: Date;
+  name: string;
+  sub_project: string | null;
```
suyashpatil78, line 7: `null` is not required here!

src/app/core/models/platform/platform-project.model.ts
```diff
@@ -0,0 +1,13 @@
...
+  code: string | null;
```
suyashpatil78, line 8: same as above

src/app/core/services/projects.service.ts
suyashpatil78: yes sire!

src/app/core/models/platform/platform-project.model.ts
SahilK-027: OK

SahilK-027:
There are multiple conflicts I will create a new separate PR for this.

PR #2923, opened 2024-04-30
feat: Replace ExtendedProject -> ProjectV2

SahilK-027:
### Description
While migrating ProjectsAPI to Platform we have noticed that we are using three models
- More Specifically:
  - v1 - ProjectV1
  - v2 - ExtendedProject
  - platform - ProjectPlatform
  
The naming convention `ExtendedProject` looks incorrect. If we see the project service method, it uses `/v2 API `so the model name would be `ProjectV2` .
So we have decided we will be renaming and replacing

ExtendedProject -> ProjectV2 . Across the entire mobile-app wherever we are using it.
```diff
- ExtendedProject
+ ProjectV2
```

## Clickup
[Please add link here](https://app.clickup.com/t/86cv8cpkt)


suyashpatil78 (approved):

src/app/core/mock-data/recently-used.data.ts
```diff
@@ -1,5 +1,5 @@
 import { RecentlyUsed } from '../models/v1/recently_used.model';
-import { ExtendedProject } from '../models/v2/extended-project.model';
+import { ProjectV2 } from '../models/v2/extended-project.model';
```
Dimple16, line 2: we should change the file name from `extended-project.model` to something like -> `project-v2.model`

Dimple16 (approved):
one comment

src/app/core/mock-data/recently-used.data.ts
SahilK-027: Yes I missed it 👍, will change.

PR #2946, opened 2024-05-03
feat: Migrate getbyId -> spenderPlatformV1ApiService

SahilK-027:
### Description
Successfully migrated `getbyId` -> `spenderPlatformV1ApiService`
Tested working for ✅:
```js
add-edit-advance.page
add-edit-expense.page
add-edit-mileage.page
add-edit-per-diem.page
fy-select-project-modal.component
```

Screenshot 2024-05-03 at 5 30 42 PM

## Clickup
[Please add link here](https://app.clickup.com/t/86cv9xc5e)


src/app/core/services/projects.service.spec.ts
```diff
@@ -44,27 +48,35 @@ describe('ProjectsService', () => {
...
     expect(projectsService).toBeTruthy();
   });
 
   it('should be able to fetch project by id', (done) => {
-    apiV2Service.get.and.returnValue(of(apiV2ResponseSingle));
+    spenderPlatformV1ApiService.get.and.returnValue(of(platformProjectSingleRes));
+
     projectsService.getbyId(257528).subscribe((res) => {
-      expect(res).toEqual(fixDate(apiV2ResponseSingle.data)[0]);
+      expect(apiV2ResponseSingle.data[0]).toEqual(jasmine.objectContaining(res));
```
devendrafyle, line 73: is it the right way to test mocked objects?

devendrafyle (approved):
Minor comment

src/app/core/services/projects.service.spec.ts
SahilK-027: Mock data has some fields that are not there in the `PlatformAPI` response so I am testing it against `objectContaining`. I think this is the correct way to test otherwise we will have to change the entire APIV2 mock data.

devendrafyle: create new mock data, sir, that depicts the real scenario after the update, why apiV2Data if we are hitting platform API

suyashpatil78: +1

SahilK-027: Sure @devendrafyle 👍

src/app/core/services/projects.service.spec.ts
```diff
@@ -44,27 +48,35 @@ describe('ProjectsService', () => {
...
+    spenderPlatformV1ApiService.get.and.returnValue(of(platformProjectSingleRes));
+
     projectsService.getbyId(257528).subscribe((res) => {
-      expect(res).toEqual(fixDate(apiV2ResponseSingle.data)[0]);
+      expect(res).toEqual(apiV2ResponseSingle.data[0]);
       done();
     });
 
-    expect(apiV2Service.get).toHaveBeenCalledWith('/projects', {
+    expect(spenderPlatformV1ApiService.get).toHaveBeenCalledWith('/projects', {
```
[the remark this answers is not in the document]
SahilK-027: Done sir

src/app/core/services/projects.service.spec.ts
```diff
@@ -44,29 +48,36 @@ describe('ProjectsService', () => {
...
   it('should be able to fetch project by id', (done) => {
-    apiV2Service.get.and.returnValue(of(apiV2ResponseSingle));
+    spenderPlatformV1ApiService.get.and.returnValue(of(platformProjectSingleRes));
+
     projectsService.getbyId(257528).subscribe((res) => {
-      expect(res).toEqual(fixDate(apiV2ResponseSingle.data)[0]);
+      expect(res).toEqual(apiV2ResponseSingle.data[0]);
+      expect(spenderPlatformV1ApiService.get).toHaveBeenCalledOnceWith('/projects', {
+        params: {
+          id: 'eq.257528',
```
devendrafyle, line 76: why the `transform` method is not mocked? please test that function as well if it's being called or used.

SahilK-027: Done!

src/app/core/test-data/projects.spec.data.ts
```diff
@@ -1,5 +1,6 @@
+import { PlatformApiResponse } from '../models/platform/platform-api-response.model';
+import { PlatformProject } from '../models/platform/platform-project.model';
```
Dimple16, line 2: we shouldn't be adding the mock data here
there should be a mock data file for platformProject in the mock data folder 
`/mock-data/platform-project.data.ts` 
All the mock data related to `PlatformProject` model should be added there

src/app/core/test-data/projects.spec.data.ts
```diff
@@ -170,6 +165,26 @@ export const apiV2ResponseSingle = deepFreeze({
   url: '/v2/projects',
 });
 
+export const platformProjectSingleRes: PlatformApiResponse<PlatformProject[]> = deepFreeze({
```
Dimple16, line 168: this should be in a separate file

Dimple16 (changes requested):
added comments

src/app/core/test-data/projects.spec.data.ts
SahilK-027: Ok

src/app/core/test-data/projects.spec.data.ts
SahilK-027: Ok

src/app/core/test-data/platformProjects.spec.data.ts
```diff
@@ -0,0 +1,23 @@
+import { PlatformApiResponse } from '../models/platform/platform-api-response.model';
+import { PlatformProject } from '../models/platform/platform-project.model';
+import deepFreeze from 'deep-freeze-strict';
```
Dimple16, line 3: why is this needed?

Dimple16 (changes requested):
one question
we shouldn't be using `deepFreeze` 
Why is it needed?

src/app/core/test-data/platformProjects.spec.data.ts
SahilK-027: @suyashpatil78 can you please explain?

SahilK-027: - If there is a common global mock data that we are using in two different places, then if the data gets modified in one place it affects the other test cases in different files.
- So we need to create a deep copy of data but it's not feasible to make a deep copy in every place we are using and modifying the mock data, hence we are using a new dependency `deep-freeze`.

Dimple16 (approved):

PR #2953, opened 2024-05-07
feat: Move Create report to platform [5]

bistaastha:
PR content:
- Moves Create Report to platform
- Removes old create report dialog - this is not in use anymore and was causing issues while removing the createDraft method from reports service


src/app/core/services/platform/v1/spender/reports.service.spec.ts
```diff
@@ -232,6 +233,7 @@ describe('SpenderReportsService', () => {
     spenderReportsService.createDraft(reportParam).subscribe((res) => {
       expect(res).toEqual(allReportsPaginated1.data[0]);
       expect(spenderPlatformV1ApiService.post).toHaveBeenCalledOnceWith('/reports', reportParam);
+      spyOn(spenderReportsService, 'clearTransactionCache').and.returnValue(of(null));
```
bistaastha, line 236: Fixed in follow up

src/app/core/services/platform/v1/spender/reports.service.ts
```diff
@@ -73,6 +73,15 @@ export class SpenderReportsService {
     );
   }
 
+  @CacheBuster({
+    cacheBusterNotifier: reportsCacheBuster$,
+  })
+  createDraft(data: CreateDraftParams): Observable<Report> {
+    return this.spenderPlatformV1ApiService
+      .post<PlatformApiPayload<Report>>('/reports', data)
+      .pipe(switchMap((res) => this.clearTransactionCache().pipe(map(() => res.data))));
```
Chethan-Fyle, line 82: Why nested pipe?

Chethan-Fyle: Use `tap()` if required.

Chethan-Fyle: The method name is `createDraft()` but I don't see anything in the definition that relates to draft. I see only the creation of a regular report.

src/app/core/services/report.service.ts
```diff
@@ -130,18 +132,9 @@ export class ReportService {
...
-  }
-
-  @CacheBuster({
-    cacheBusterNotifier: reportsCacheBuster$,
-  })
-  create(report: ReportPurpose, expenseIds: string[]): Observable<ReportV1> {
-    return this.createDraft(report).pipe(
-      switchMap((newReport: ReportV1) => {
+  create(report: ReportPurpose, expenseIds: string[]): Observable<Report> {
+    return this.spenderReportsService.createDraft({ data: report }).pipe(
```
Chethan-Fyle, line 136: There we have create() method and inside we are calling `createDraft()`. This can be refactored IMO.

Chethan-Fyle (changes requested):
Few comments.

mvaishnavi (approved):
another review from me not needed.

src/app/core/services/platform/v1/spender/reports.service.ts
bistaastha: This method creates draft reports, for a submitted report this API is used in combination with `submit` API from public

Chethan-Fyle (approved):

PR #2967, opened 2024-05-09
feat: Migrate getAllActive -> spenderPlatformV1ApiService

SahilK-027:
## Clickup
[Please add link here](https://app.clickup.com/t/86cv9n9gt)


Dimple16 (approved):

src/app/core/services/projects.service.spec.ts
```diff
@@ -83,17 +86,18 @@ describe('ProjectsService', () => {
   });
 
   it('should be able to fetch all active projects', (done) => {
-    apiService.get.and.returnValue(of(apiResponseActiveOnly));
+    spenderPlatformV1ApiService.get.and.returnValue(of(platformAPIResponseActiveOnly));
+    const transformToV1ResponseSpy = spyOn(projectsService, 'transformToV1Response').and.callThrough();
```
devendrafyle, line 90: Could you please check how we are testing the `transform` method in platform expense APIs, in that case, we are testing it from the `transaction` service the only change in your case is that you have to test it from the project service itself. 

By testing that way we not only ensure that the `transform` method is called, but also we test for what input what output to expect.

devendrafyle (approved):
minor comment

src/app/core/services/projects.service.spec.ts
suyashpatil78: +1

src/app/core/services/projects.service.spec.ts
```diff
@@ -83,17 +86,18 @@ describe('ProjectsService', () => {
...
     projectsService.getAllActive().subscribe((res) => {
       expect(res).toEqual(expectedReponseActiveOnly);
+      expect(spenderPlatformV1ApiService.get).toHaveBeenCalledWith('/projects', {
```
suyashpatil78, line 93: `toHaveBeenCalledOnceWith`

suyashpatil78 (commented):
2 comments

src/app/core/services/projects.service.spec.ts
SahilK-027: OK

src/app/core/services/projects.service.spec.ts
SahilK-027: Yup I will change it 👍

src/app/core/services/projects.service.spec.ts
SahilK-027: Will work on this in separate PR for all the methods at once.

cc: @devendrafyle , @suyashpatil78

suyashpatil78 (approved):

PR #3028, opened 2024-05-28
fix: fix test for permission service

bistaastha:


src/app/core/services/permissions.service.spec.ts
```diff
@@ -30,7 +30,7 @@ describe('PermissionsService', () => {
 
   describe('allowedActions():', () => {
     it('should get allowed actions for a user', (done) => {
-      const roles = ['HOP', 'HOD', 'OWNER'];
```
bistaastha, line 33: these changes have been made to accomodate for changes in `reports.approve` flag to false for hod and hop roles

PR #3053, opened 2024-06-04
fix: breaking edit expense page

SahilK-027:
### Description
Why map caused the error and will concatMap work?
In this case, this.activeCategories$ emits categories and this.projectsService.getbyId()  returns an Observable representing the project data.
- `map`:  it would try to emit the Observable returned by getbyId directly. This would result in an error because the subscribe method expects a single value, not another Observable.
- `concatMap`: By using concatMap, we are instructing it to subscribe to the inner Observable (from getbyId) and then merge the project data emissions into the final stream.

## Clickup
[click-up](https://app.clickup.com/t/86cvjz2jy)

### DIFF
--- a/src/app/fyle/add-edit-expense/add-edit-expense.page.ts
+++ b/src/app/fyle/add-edit-expense/add-edit-expense.page.ts
@@ -2542,11 +2542,11 @@ export class AddEditExpensePage implements OnInit {
       switchMap((etxn) => {
         if (etxn.tx.project_id) {
           return this.activeCategories$.pipe(
-            map((allActiveCategories) => this.projectsService.getbyId(etxn.tx.project_id, allActiveCategories))
+            concatMap((allActiveCategories) => this.projectsService.getbyId(etxn.tx.project_id, allActiveCategories))
           );
         } else if (projectControl?.value?.project_id) {
           return this.activeCategories$.pipe(
-            map((allActiveCategories) =>
+            concatMap((allActiveCategories) =>
               this.projectsService.getbyId(projectControl.value.project_id, allActiveCategories)
             )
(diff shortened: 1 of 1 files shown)


devendrafyle (approved):

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
```diff
@@ -2542,11 +2542,11 @@ export class AddEditExpensePage implements OnInit {
       switchMap((etxn) => {
         if (etxn.tx.project_id) {
           return this.activeCategories$.pipe(
-            map((allActiveCategories) => this.projectsService.getbyId(etxn.tx.project_id, allActiveCategories))
+            concatMap((allActiveCategories) => this.projectsService.getbyId(etxn.tx.project_id, allActiveCategories))
```
Chethan-Fyle, line 2545: Does `activeCategories$` return multiple values over a period of time? If not, `concatMap` is not the suitable operator, even though it works.
If `activeCategories$` return single value, use `switchMap` operator.

Chethan-Fyle (commented):
One comment.

src/app/fyle/add-edit-expense/add-edit-expense.page.ts
SahilK-027: Yes `switchMap` will work.

To avoid nested `switchMap` I am using combineLatest.

SahilK-027: Done 👍

Dimple16 (approved):

suyashpatil78 (approved):

PR #3151, opened 2024-07-17
fix: fix empty search string returning no projects

bistaastha:

### DIFF
--- a/src/app/core/services/projects.service.ts
+++ b/src/app/core/services/projects.service.ts
@@ -82,7 +82,7 @@ export class ProjectsService {
   }
 
   addNameSearchFilter(searchNameText: string, params: PlatformProjectParams): void {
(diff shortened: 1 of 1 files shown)


src/app/core/services/projects.service.ts
```diff
@@ -82,7 +82,7 @@ export class ProjectsService {
   }
 
   addNameSearchFilter(searchNameText: string, params: PlatformProjectParams): void {
-    if (typeof searchNameText !== 'undefined' && searchNameText !== null) {
+    if (typeof searchNameText !== 'undefined' && searchNameText !== null && searchNameText !== '') {
```
suyashpatil78, line 85: `if (typeof searchNameText !== 'undefined' && searchNameText)`

suyashpatil78 (approved):
One minor comment

Chethan-Fyle (approved):

PR #3162, opened 2024-08-05
feat: Added deep link redirection support for sms opt in

arjunaj5:
### Description
copilot:summary

copilot:poem

### Walkthrough
copilot:walkthrough

## Clickup
Please add link here

## Code Coverage
Please add code coverage here

## UI Preview

### DIFF
--- a/src/app/core/services/deep-link.service.ts
+++ b/src/app/core/services/deep-link.service.ts
@@ -3,12 +3,19 @@ import { Router } from '@angular/router';
 import { Redirect } from '../models/redirect.model';
 import { UnflattenedTransaction } from '../models/unflattened-transaction.model';
 import { TrackingService } from './tracking.service';
+import { OrgService } from './org.service';
+import { ModalController } from '@ionic/angular';
 
 @Injectable({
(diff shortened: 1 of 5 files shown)


arjunaj5:
Will be covering the tests in a sub PR

src/app/core/services/deep-link.service.ts
```diff
@@ -109,6 +111,15 @@ export class DeepLinkService {
           orgId,
           txnId,
         });
+      } else if (redirectUri.match('dashboard') && openSMSOptInDialog === 'true') {
```
suyashpatil78, line 114: can we be more specific, if in future we have more such instances that opens in dashboard, this will cause conflicts

suyashpatil78 (approved):

src/app/core/services/deep-link.service.ts
arjunaj5: we are also checking for `open_sms_dialog` present in the deep link no? so isn't this specific enough?

Chethan-Fyle (approved):

PR #3167, opened 2024-08-07
fix: replace cached EOU object with an API call on settings page

harshal015:
### Description
copilot:summary

copilot:poem

### Walkthrough
copilot:walkthrough

## Clickup
Please add link here
https://app.clickup.com/t/86cw56hkr

## Code Coverage
Please add code coverage here

## UI Preview

### DIFF
--- a/src/app/fyle/my-profile/my-profile.page.ts
+++ b/src/app/fyle/my-profile/my-profile.page.ts
@@ -181,7 +181,7 @@ export class MyProfilePage {
 
   ionViewWillEnter(): void {
     this.setupNetworkWatcher();
-    this.eou$ = from(this.authService.getEou());
+    this.eou$ = this.authService.refreshEou();
     this.isUserFromINCluster$ = from(this.utilityService.isUserFromINCluster());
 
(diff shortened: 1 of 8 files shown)


src/app/fyle/my-profile/my-profile.page.ts
```diff
@@ -181,7 +181,7 @@ export class MyProfilePage {
 
   ionViewWillEnter(): void {
     this.setupNetworkWatcher();
-    this.eou$ = from(this.authService.getEou());
+    this.eou$ = this.authService.refreshEou();
```
suyashpatil78, line 184: every time we visit settings page, this call will be made which is kind of unnecessary. Can we check if we can refresh every time we open the app?

suyashpatil78: this will also help us avoid the opt-in nudges in dashboard if some already has opted in.

suyashpatil78 (changes requested):
1 question

suyashpatil78 (approved):

PR #3171, opened 2024-08-14
fix: complete method typeerror

Madhusudan092001:
## Clickup
app.clickup.com

## Code Coverage
TypeError
s.complete is not a function. (In 's.complete()', 's.complete' is undefined)

The error `s.complete() is not a function` occurs when the `complete` method is either undefined or not a function on the target object. I confirmed through console logs that `event.target` is a valid `ion-infinite-scroll` element, and the `complete` method exists as a function. However, if the method doesn't exist for a specific user, it might be missing or incorrectly bound to the target element. As a fallback, I implemented a check to ensure that if `event.target` or `complete` isn't a function, the `complete` method won't be called, preventing the error.

## UI Preview

### DIFF
--- a/src/app/fyle/team-reports/team-reports.page.ts
+++ b/src/app/fyle/team-reports/team-reports.page.ts
@@ -252,7 +252,11 @@ export class TeamReportsPage implements OnInit {
     params.pageNumber = this.currentPageNumber;
     this.loadData$.next(params);
     setTimeout(() => {
-      event?.target?.complete();
+      if (typeof event?.target?.complete === 'function') {
+        event.target.complete();
+      } else {
+        console.error('complete() method is not available on the target element');
+      }
     }, 1000);
   }
 
--- a/src/app/fyle/team-reports/team-reports.page.ts
+++ b/src/app/fyle/team-reports/team-reports.page.ts
@@ -252,11 +252,7 @@ export class TeamReportsPage implements OnInit {
     params.pageNumber = this.currentPageNumber;
     this.loadData$.next(params);
     setTimeout(() => {
-      if (typeof event?.target?.complete === 'function') {
-        event.target.complete();
-      } else {
-        console.error('complete() method is not available on the target element');
-      }
+      event?.target?.complete?.();
     }, 1000);
   }
 


src/app/fyle/team-reports/team-reports.page.ts
```diff
@@ -252,7 +252,11 @@ export class TeamReportsPage implements OnInit {
     params.pageNumber = this.currentPageNumber;
     this.loadData$.next(params);
     setTimeout(() => {
-      event?.target?.complete();
+      if (typeof event?.target?.complete === 'function') {
```
suyashpatil78, line 255: can we use JavaScript optional chaining here?

suyashpatil78 (changes requested):

src/app/fyle/team-reports/team-reports.page.ts
Madhusudan092001: Yes sir we can also use optional chaining here as it will simplifies the code as compared to the typeof check
I guess we can do it like this :

if (!event?.target?.complete) { 
  console.error('complete() method is not available on the target element');
} else {
  event.target.complete();
}

Sir here using the optional chaining ?.  the if block (!event?.target?.complete) will execute when either of the following happens :
event is null or undefined
event.target is null or undefined
event.target.complete is null, undefined, or false

So we are not directly calling the complete method when it can became undefined or not available which leads to the typeerror

And when complete is a function then we can directly access it in the else using event.target.complete();

suyashpatil78: We don't push console logs/errors to production

suyashpatil78: Can we use optional chaining to check if `complete` method exist or not on `target` object?

Madhusudan092001: Yes sir i guess then we can try like this without consoling:

Sir the optional chaining will work in a similar manner as above for checking in event?.target?.complete(); 

const completed = event?.target?.complete(); //Here we can assign the result of optional chaining into a variable

//And now if it gives a true value then only we are accessing it or calling, and if undefined then it will 
                               be falsy and not called
if (completed) { 
  completed();
}

suyashpatil78: Can you think of another approach? We can use optional chaining for the complete method also.

Madhusudan092001: Yes sir then we can directly check using the optional chaining without the different if else or variable declarations

Like  event?.target?.complete?.();
It will check complete method exist on target or not and if not it will evaluate to undefined without throwing an error

suyashpatil78: Nice! This will be cleaner and works well.

suyashpatil78 (approved):

PR #3172, opened 2024-08-15
fix: Replace public implementation to display delegatee info with platform 

harshal015:
## Clickup
Please add link here
app.clickup.com
## Code Coverage
Please add code coverage here

## UI Preview

### DIFF
--- a/src/app/fyle/delegated-accounts/delegated-accounts.page.html
+++ b/src/app/fyle/delegated-accounts/delegated-accounts.page.html
@@ -26,11 +26,11 @@
         <ion-grid>
           <ion-row>
             <ion-col size="2">
-              <div class="delegated--org-thumb">{{ acc?.full_name | initials | uppercase }}</div>
+              <div class="delegated--org-thumb">{{ acc.full_name | initials | uppercase }}</div>
             </ion-col>
             <ion-col size="10">
-              <div class="delegated--org-element-name">{{acc?.full_name}}</div>
-              <div class="delegated--org-element-us-details">{{acc?.email}}</div>
+              <div class="delegated--org-element-name">{{acc.full_name}}</div>
+              <div class="delegated--org-element-us-details">{{acc.email}}</div>
               <div class="delegated--org-element-us-details">{{currentOrg.name}}</div>
             </ion-col>
           </ion-row>
--- a/src/app/fyle/delegated-accounts/delegated-accounts.page.html
+++ b/src/app/fyle/delegated-accounts/delegated-accounts.page.html
@@ -26,11 +26,11 @@
         <ion-grid>
           <ion-row>
             <ion-col size="2">
-              <div class="delegated--org-thumb">{{ acc.full_name | initials | uppercase }}</div>
+              <div class="delegated--org-thumb">{{ acc?.full_name | initials | uppercase }}</div>
             </ion-col>
             <ion-col size="10">
-              <div class="delegated--org-element-name">{{acc.full_name}}</div>
-              <div class="delegated--org-element-us-details">{{acc.email}}</div>
+              <div class="delegated--org-element-name">{{acc?.full_name}}</div>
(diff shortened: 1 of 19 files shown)


src/app/fyle/delegated-accounts/delegated-accounts.page.html
```diff
@@ -31,9 +31,7 @@
             <ion-col size="10">
               <div class="delegated--org-element-name">{{acc?.us?.full_name}}</div>
               <div class="delegated--org-element-us-details">{{acc?.us?.email}}</div>
-              <div *ngIf="acc?.ou.org_id === currentOrg.id" class="delegated--org-element-us-details">
-                {{currentOrg.name}}
-              </div>
+              <div class="delegated--org-element-us-details">{{currentOrg.name}}</div>
```
mvaishnavi, line 34: why have we removed the condition here? @harshal015

harshal015: it'll be anyway visible for all delegators as the API will respond with delegators from same org.

harshal015: can entirely remove the org name as well

src/app/fyle/notifications/notifications.page.ts
```diff
@@ -282,7 +284,10 @@ export class NotificationsPage implements OnInit {
       this.notificationForm.controls.notifyOption.setValue(notifyOption);
     });
 
-    this.isDelegateePresent$ = from(this.authService.getEou()).pipe(map((eou) => eou.ou.delegatee_id !== null));
+    this.isDelegateePresent$ = from(this.authService.getEou()).pipe(
+      switchMap((res) => this.employeesService.getByParams({ user_id: `eq.${res.us.id}` })),
+      map((employee) => employee.data.delegatees.length > 0)
```
suyashpatil78, line 289: If delegatees is null this would break right?
We have mentioned delegatees as optional in model and also as per docs.fylehq.com it could be null.
Use optional chaining.

src/app/fyle/notifications/notifications.page.spec.ts
```diff
@@ -95,6 +103,7 @@ describe('NotificationsPage', () => {
     });
 
     orgUserSettingsService.get.and.returnValue(of(orgUserSettingsData));
+    employeesService.getByParams.and.returnValue(of());
```
suyashpatil78, line 106: Why are we not returning anything?
Either it should return employees or throw an error

src/app/fyle/delegated-accounts/delegated-accounts.page.html
```diff
@@ -26,14 +26,12 @@
...
-              <div class="delegated--org-thumb">{{ acc?.us?.full_name | initials | uppercase }}</div>
+              <div class="delegated--org-thumb">{{ acc?.full_name | initials | uppercase }}</div>
             </ion-col>
             <ion-col size="10">
-              <div class="delegated--org-element-name">{{acc?.us?.full_name}}</div>
-              <div class="delegated--org-element-us-details">{{acc?.us?.email}}</div>
-              <div *ngIf="acc?.ou.org_id === currentOrg.id" class="delegated--org-element-us-details">
-                {{currentOrg.name}}
-              </div>
+              <div class="delegated--org-element-name">{{acc?.full_name}}</div>
```
suyashpatil78, line 32: What happens if acc is undefined or null?
Are we handling the null cases as well?

suyashpatil78 (changes requested):

src/app/fyle/delegated-accounts/delegated-accounts.page.html
harshal015: we are displaying switch accounts page only if its non empty, so it'll always have a value

suyashpatil78 (approved):

PR #3204, opened 2024-09-12
fix: Add optional chaining to check if id exists before redirection

bistaastha:
Sentry issue: https://fyle-technologies-private-limi.sentry.io/issues/5770926111/?project=5622998&referrer=ClickUp

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved navigation stability by adding checks for the existence of `id` before routing, preventing potential runtime errors.

### DIFF
--- a/src/app/fyle/dashboard/tasks/tasks.component.ts
+++ b/src/app/fyle/dashboard/tasks/tasks.component.ts
@@ -508,7 +508,9 @@ export class TasksComponent implements OnInit {
           finalize(() => this.loaderService.hideLoader())
         )
         .subscribe((res) => {
-          this.router.navigate(['/', 'enterprise', 'my_view_report', { id: res[0].id }]);
+          if (res[0]?.id) {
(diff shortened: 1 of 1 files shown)


src/app/fyle/dashboard/tasks/tasks.component.ts
```diff
@@ -508,7 +508,9 @@ export class TasksComponent implements OnInit {
           finalize(() => this.loaderService.hideLoader())
         )
         .subscribe((res) => {
-          this.router.navigate(['/', 'enterprise', 'my_view_report', { id: res[0].id }]);
+          if (res[0]?.id) {
```
suyashpatil78, line 511: Should we show error toast in case `id` is not present?

bistaastha: No, I don't think that is necessary, there are already checks in place to ensure that the task is only shown in cases where there is a valid report/advance. This is just for an edge case.

PR #3215, opened 2024-09-24
fix: fix reference for system_category field to make sub categories visible in Mileage Form

bistaastha:
## ClickUp
https://app.clickup.com/t/86cwk7q7w

## Description
The sub category field was not visible in mobile app Mileage form, because the sub category filter was returning null. Instead of checking `name`, the system category needs to be checked here.

The web app equivalent for this filtering logic is getting category by checking if system category is mileage and sub category not being null, as follows:
```
is_enabled: eq.true
limit: 200
offset: 0
sub_category: neq.null
system_category: eq.Mileage
```

## Summary by CodeRabbit

- **New Features**
	- Enhanced filtering logic for mileage and per diem categories, improving the accuracy of displayed options.
	- Updated methods now return both primary and secondary mileage categories, ensuring users have access to all relevant data.

- **Bug Fixes**
	- Corrected categorization of mileage data to ensure proper alignment with intended classifications.

- **Tests**
	- Adjusted test cases to reflect changes in category filtering and expected outputs, ensuring reliability of functionality.

### DIFF
--- a/src/app/fyle/add-edit-mileage/add-edit-mileage.page.ts
+++ b/src/app/fyle/add-edit-mileage/add-edit-mileage.page.ts
@@ -671,7 +671,7 @@ export class AddEditMileagePage implements OnInit {
         const parentCategoryName = 'mileage';
         return categories.filter(
           (orgCategory) =>
-            parentCategoryName.toLowerCase() === orgCategory.name?.toLowerCase() &&
+            parentCategoryName.toLowerCase() === orgCategory.fyle_category?.toLowerCase() &&
             parentCategoryName.toLowerCase() !== orgCategory.sub_category?.toLowerCase()
         );
       }),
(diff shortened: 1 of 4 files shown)


src/app/fyle/add-edit-mileage/add-edit-mileage.page.ts
```diff
@@ -671,7 +671,7 @@ export class AddEditMileagePage implements OnInit {
         const parentCategoryName = 'mileage';
         return categories.filter(
           (orgCategory) =>
-            parentCategoryName.toLowerCase() === orgCategory.name?.toLowerCase() &&
+            parentCategoryName.toLowerCase() === orgCategory.fyle_category?.toLowerCase() &&
```
Dimple16, line 674: We should fix this for per diem expense too

Dimple16 (approved):
Please fix per diem form along with this before merging

Chethan-Fyle (approved):

suyashpatil78 (approved):

bistaastha:
Updated test references, no additional logic added, changes working as expected on staging

PR #3220, opened 2024-10-09
feat: migrate public currency and exchange rate api to platform 

ravikumar-hash:
Clickup
https://app.clickup.com/t/86cwkequh

Description:
1. Added platformCommonApiService. 
2. Migrated public Currency and exchange rate Api's to platform
3. Added tests for services. 

## Summary by CodeRabbit

- **New Features**
	- Updated currency data with new entries and modifications, including the addition of 'Chinese Yuan Offshore' and updates to existing currency codes and names.
	- Introduced a new interface for structured exchange rate data representation.
	- Added a new service for improved API interactions.

- **Bug Fixes**
	- Adjusted API endpoints and query parameters for fetching exchange rates and currency lists to ensure accurate data retrieval.

- **Tests**
	- Updated test suites for `CurrencyService` and `RouterAuthService` to reflect changes in dependencies and ensure proper functionality.

### DIFF
--- a/src/app/core/services/currency.service.ts
+++ b/src/app/core/services/currency.service.ts
@@ -6,7 +6,7 @@ import { catchError, map, switchMap } from 'rxjs/operators';
 import { Cacheable } from 'ts-cacheable';
 import { CurrencyName } from '../models/currency.model';
 import { ExtendedOrgUser } from '../models/extended-org-user.model';
-import { PublicPolicyExpense } from '../models/public-policy-expense.model';
+import { ExchangeRateResponse } from '../models/exchange-rate-response.model';
 import { AuthService } from './auth.service';
 import { OrgService } from './org.service';
 import { PlatformCommonApiService } from './platform-common-api.service';
@@ -35,7 +35,7 @@ export class CurrencyService {
     }
 
     return this.platformCommonApiService
-      .get<PlatformApiResponse<Partial<PublicPolicyExpense>>>('/currency/exchange_rate', {
+      .get<PlatformApiResponse<ExchangeRateResponse>>('/currency/exchange_rate', {
         params: queryParams,
       })
       .pipe(
--- a/src/app/core/services/currency.service.ts
+++ b/src/app/core/services/currency.service.ts
@@ -51,7 +51,7 @@ export class CurrencyService {
         this.platformCommonApiService
           .get<PlatformApiResponse<CurrencyName>>('/currency/list', {
             params: {
-              org_id: currentEou && currentEou.ou && currentEou.ou.org_id,
+              org_id: currentEou?.ou?.org_id,
             },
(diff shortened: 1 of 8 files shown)


src/app/core/services/currency.service.ts
```diff
@@ -44,11 +48,13 @@ export class CurrencyService {
...
       switchMap((currentEou: ExtendedOrgUser) =>
-        this.apiService.get<CurrencyName>('/currency/all', {
-          params: {
-            org_id: currentEou && currentEou.ou && currentEou.ou.org_id,
-          },
-        })
+        this.platformCommonApiService
+          .get<PlatformApiResponse<CurrencyName>>('/currency/list', {
+            params: {
+              org_id: currentEou && currentEou.ou && currentEou.ou.org_id,
```
Dimple16, line 54: `org_id: currentEou?.ou?.org_id` this should work

src/app/core/services/currency.service.ts
```diff
@@ -44,11 +48,13 @@ export class CurrencyService {
...
+            },
+          })
+          .pipe(map((response) => response.data))
```
Dimple16, line 57:
```
return from(this.authService.getEou()).pipe(
    switchMap((currentEou: ExtendedOrgUser) => {
      return this.platformCommonApiService.get<PlatformApiResponse<CurrencyName[]>>('/currency/list', {
        params: { org_id: currentEou?.ou?.org_id },
      });
    }),
    map((response) => response.data)
  );
```
I don't think another `pipe` is required here. Check if the above works

Dimple16 (changes requested):
added comments

src/app/core/services/currency.service.ts
ravikumar-hash: I tried  this method but it was not working, so i searched and found that, , map is applied after the switchMap, switchMap returns is an observable from the HTTP call, we want  map() to act on the response within the switchMap function. By placing map() inside the pipe(), we can correctly get the response data.

Dimple16 (approved):

Dimple16 (changes requested):
@ravikumar-hash , can you check the unit tests?

src/app/core/services/platform-common-api.service.ts
```diff
@@ -0,0 +1,23 @@
+import { HttpClient } from '@angular/common/http';
+import { Injectable } from '@angular/core';
+import { Observable } from 'rxjs/internal/Observable';
+import { environment } from 'src/environments/environment';
```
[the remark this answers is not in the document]
Dimple16: @ravikumar-hash we should be using this `import { Observable } from 'rxjs';`

Dimple16 (approved):
added a comment

PR #3250, opened 2024-11-04
feat: Removing url query params from Http exception for correct grouping of issues in Sentry

Aniruddha-Shriwant:
I have added new fingerprint rule in Sentry: https://fyle-technologies-private-limi.sentry.io/settings/projects/fyle-mobile-app-prod/issue-grouping/

```
# New issue should be created for unique API endpoint
error.value:"Http failure response for*" -> http-error, {{error.value}}
```
this will create grouping of issues based on api endpoints for all http failure errors, but url query params were messing up this grouping hence removing them from the exception value - they would be still visible in stack trace

## Clickup
https://app.clickup.com/t/86cwy4tvn

Webapp PR: https://github.com/fylein/fyle-app/pull/8921

## Summary by CodeRabbit

- **New Features**
	- Enhanced error reporting for HTTP exceptions by excluding query parameters from URLs before sending to Sentry.
- **Bug Fixes**
	- Improved clarity and grouping of similar HTTP errors in error tracking.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -16,11 +16,12 @@ GlobalCacheConfig.maxCacheCount = 100;
 
 // To modify the exception values for Http errors to remove query params from the URL so that grouping is done properly in Sentry
 const cleanHttpExceptionUrlsForSentryGrouping = (event: Sentry.Event): void => {
+  // exceptionValue looks like: 'Http failure response for https://staging.fyle.tech/platform/v1/common/currency/exchange_rate?from=null&to=USD&date=2024-10-01: 400 OK'
+  const exceptionValue = event.exception?.values?.[0].value;
   const prefix = 'Http failure response for ';
   const suffixIndicator = ': ';
-  if (event.exception?.values?.[0].value?.startsWith(prefix)) {
-    // exceptionValue looks like: 'Http failure response for https://staging.fyle.tech/platform/v1/common/currency/exchange_rate?from=null&to=USD&date=2024-10-01: 400 OK'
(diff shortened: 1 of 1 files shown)


arjunaj5 (approved):

src/main.ts
```diff
@@ -14,12 +14,35 @@ import { GlobalCacheConfig } from 'ts-cacheable';
 GlobalCacheConfig.maxAge = 10 * 60 * 1000;
 GlobalCacheConfig.maxCacheCount = 100;
 
+// To modify the exception values for Http errors to remove query params from the URL so that grouping is done properly in Sentry
+const cleanHttpExceptionUrlsForSentryGrouping = (event: Sentry.Event): void => {
+  const prefix = 'Http failure response for ';
+  const suffixIndicator = ': ';
+  if (event.exception?.values?.[0].value?.startsWith(prefix)) {
+    // exceptionValue looks like: 'Http failure response for https://staging.fyle.tech/platform/v1/common/currency/exchange_rate?from=null&to=USD&date=2024-10-01: 400 OK'
+    const exceptionValue = event.exception?.values?.[0].value;
```
Chethan-Fyle, line 23: Move this outside of enclosing `if()` block so that it can be reused in the if condition as well.

Chethan-Fyle (approved):
One minor comment.

src/main.ts
Aniruddha-Shriwant: ahh correct, will do!

PR #3438, opened 2025-01-23
fix: Uses guard instead of redirection to reach onboarding page when opening the mobile app

bistaastha:
This PR includes:
- Changing the existing implementation for redirection to onboarding page
- Refactoring code

## Release Notes

- **New Features**
  - Introduced an Onboarding Guard to control route access based on user and organization status.
  - Enhanced navigation logic to dynamically determine dashboard or onboarding redirections.

- **Improvements**
  - Streamlined user navigation flow across multiple pages.
  - Simplified service methods for checking onboarding requirements.
  - Added more sophisticated checks for organization and user eligibility.

- **Changes**
  - Modified default routing to redirect users to 'my_dashboard' as the new initial landing page.
  - Updated navigation methods in multiple components to use a centralized onboarding service.

### DIFF
--- a/src/app/core/guards/onboarding.guard.ts
+++ b/src/app/core/guards/onboarding.guard.ts
@@ -1,0 +1,83 @@
+import { Injectable } from '@angular/core';
+import { ActivatedRouteSnapshot, CanActivate, Router, RouterStateSnapshot, UrlTree } from '@angular/router';
+import { forkJoin, Observable, of } from 'rxjs';
+import { map } from 'rxjs/operators';
+import { OrgSettingsService } from '../services/org-settings.service';
+import { SpenderOnboardingService } from '../services/spender-onboarding.service';
+import { UtilityService } from '../services/utility.service';
+import { OnboardingStatus } from '../models/onboarding-status.model';
+import { OrgSettings } from '../models/org-settings.model';
+import { AuthService } from '../services/auth.service';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class OnboardingGuard implements CanActivate {
+  constructor(
+    private orgSettingsService: OrgSettingsService,
+    private spenderOnboardingService: SpenderOnboardingService,
+    private utilityService: UtilityService,
+    private router: Router,
+    private authService: AuthService
+  ) {}
+
+  canActivate(
+    next: ActivatedRouteSnapshot,
+    state: RouterStateSnapshot
+  ): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
+    return forkJoin([
+      this.orgSettingsService.get(),
+      this.spenderOnboardingService.getOnboardingStatus(),
+      of(this.utilityService.isUserFromINCluster()),
+      this.authService.getEou(),
+    ]).pipe(
+      map(([orgSettings, onboardingStatus, isUserFromINCluster, eou]) => {
+        const isCCCEnabled = this.checkCCCEnabled(orgSettings);
+        const isCardFeedEnabled = this.checkCardFeedEnabled(orgSettings);
+        const restrictedOrgs = this.isRestrictedOrg(orgSettings, isUserFromINCluster);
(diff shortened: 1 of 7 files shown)


src/app/core/services/spender-onboarding.service.ts
```diff
@@ -88,4 +97,52 @@ export class SpenderOnboardingService {
...
+  checkForRedirectionToOnboarding(): Observable<boolean> {
+    return forkJoin([
+      this.orgSettingsService.get(),
+      this.getOnboardingStatus(),
+      from(this.utilityService.isUserFromINCluster()),
+      this.authService.getEou(),
+    ]).pipe(
+      map(([orgSettings, onboardingStatus, isUserFromINCluster, eou]) => {
+        const isCCCEnabled = this.checkCCCEnabled(orgSettings);
+        const isCardFeedEnabled = this.checkCardFeedEnabled(orgSettings);
```
Chethan-Fyle, line 110: isCardFeedEnabled will be true even when only amex is enabled. Isn't that wrong?

Chethan-Fyle (changes requested):
One important comment.

src/app/core/services/spender-onboarding.service.ts
bistaastha: This is intentional. If only Amex is enabled, only the second step of the onboarding flow (Opt In), will be shown to the user.

Chethan-Fyle (approved):

src/app/core/guards/onboarding.guard.ts
```diff
@@ -0,0 +1,26 @@
...
+    state: RouterStateSnapshot
+  ): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
+    return this.spenderOnboardingService.checkForRedirectionToOnboarding().pipe(
+      map((shouldRedirectToOnboarding) => {
+        if (shouldRedirectToOnboarding) {
+          return false;
+        }
+        return true;
+      })
+    );
```
arjunaj5, line 24: can we redirect to dashboard if the API throws some error?

arjunaj5, line 24: can we redirect to dashboard if the API throws some error?

src/app/fyle/fyle-routing.module.ts
```diff
@@ -2,11 +2,13 @@
 import { NgModule } from '@angular/core';
 import { Routes, RouterModule } from '@angular/router';
 import { OptInGuard } from '../core/guards/opt-in.guard';
+import { OnboardingGuard } from '../core/guards/onboarding.guard';
 
 const routes: Routes = [
   {
     path: 'my_dashboard',
     loadChildren: () => import('./dashboard/dashboard.module').then((m) => m.DashboardPageModule),
+    canActivate: [OnboardingGuard],
```
arjunaj5, line 11: might have to check if this really solves the white screen that appears for few seconds in a low network.

If it is not solved, can we land the user to the dashboard and then check if user has to go to spender onboarding flow and redirect?

arjunaj5 (commented):
2 comments

src/app/fyle/fyle-routing.module.ts
bistaastha: Going to the dashboard will still be jarring UI in this case, what I can do is try adding a loader with the guard over the white screen (same as before). The transition won't be as obvious as before though.

bistaastha: or I can add a localstorage key to determine if onboarding is complete, even before checking the API calls

src/app/core/guards/onboarding.guard.ts
bistaastha: will add this in a follow up

src/app/fyle/fyle-routing.module.ts
bistaastha: Will add this in a follow up

PR #3538, opened 2025-02-25
feat: Adding Unapproved team reports stat on dashboard

amit429:
## Clickup
https://app.clickup.com/t/86cy37ptq

## Code Coverage
Please add code coverage here

## UI Preview
image

image

## Summary by CodeRabbit

- **New Features**
	- Introduced a new method to retrieve statistics for unapproved team reports, enhancing the dashboard's reporting capabilities.
	- Added a section in the dashboard to display statistics related to unapproved team expense reports.
	- Implemented navigation to the team reports page from the unapproved team reports section.

- **Style**
	- Added a new CSS class for unapproved team reports to maintain consistent styling across report badges.


src/app/fyle/dashboard/stat-badge/stat-badge.component.html
```diff
@@ -2,7 +2,7 @@
   <div class="stat-badge--container">
     <ion-grid class="stat-badge--grid">
       <ion-row>
-        <ion-col class="stat-badge--count-col" size="3">
+        <ion-col class="stat-badge--count-col" [attr.size]="reportState === 'UNAPPROVED_TEAM_REPORTS' ? 1.5 : 3">
```
rvab, line 5: [attr.size] or [size]? Which is correct and why?

src/app/fyle/dashboard/dashboard.service.ts
```diff
@@ -46,6 +51,33 @@ export class DashboardService {
...
+          }
+          const zeroResponse: PlatformReportsStatsResponse = {
+            count: 0,
+            failed_amount: null,
+            failed_count: null,
+            processing_amount: 0,
+            processing_count: 0,
+            reimbursable_amount: 0,
+            total_amount: 0,
+          };
```
rvab, line 72: Why return this response?

src/app/fyle/dashboard/stats/stats.component.ts
```diff
@@ -134,6 +136,18 @@ export class StatsComponent implements OnInit {
...
   }
 
+  initializeUnapprovedTeamReportsStats(): void {
+    this.unapprovedTeamReportsStats$ = combineLatest({
+      currentOrg: this.orgService.getCurrentOrg(),
+      primaryOrg: this.orgService.getPrimaryOrg(),
+    }).pipe(
+      switchMap(({ currentOrg, primaryOrg }) => {
+        const showTeamReportStats = currentOrg.id === primaryOrg.id;
+        return this.dashboardService.getUnapprovedTeamReportsStats(showTeamReportStats);
```
rvab, line 146: Why call `getUnapprovedTeamReportsStats ` method if the current org is not primary org? I can see that you're doing it inside service but why not do it here?

src/app/fyle/dashboard/stats/stats.component.ts
```diff
@@ -224,6 +239,21 @@ export class StatsComponent implements OnInit {
     }
   }
 
+  goToTeamReportsPage(state: ReportStates): void {
+    if (state === ReportStates.UNAPPROVED_TEAM_REPORTS) {
+      state = ReportStates.APPROVER_PENDING;
```
rvab, line 244: Not a good practice to overwrite the `state` value like this. Assign to a different variable and update that local variable

rvab (changes requested):

src/app/fyle/dashboard/stat-badge/stat-badge.component.html
amit429: In this case [size] will be more precise as this directly to the size property of the ion-col component.
[attr.size] would bind to the size attribute of the HTML element which works but is not specific for this component.
Got it , will change this to [size]

amit429: In the recent changes the size attribute is completely gone

src/app/fyle/dashboard/dashboard.service.ts
amit429: Went through the logic once again , realised we should return null and not the zero response as its being done a bit differently for the Tasks page team reports stat logic. I took reference from there thats why added the zero response return. I have removed this and modified the code to return null

src/app/fyle/dashboard/stats/stats.component.ts
amit429: Right , I have modified the logic and taken it outside to check first is the showTeamReports is true or not and based on that I am calling the `getUnapprovedTeamReportsStats ` or returning null

rvab (approved):

PR #3547, opened 2025-02-26
fix: edge case of cost center id to add inside txnFields

Sishhhh:
## Clickup
https://app.clickup.com/t/86cy3fbf9

## Summary by CodeRabbit

- **New Features**
  - Improved expense processing with enhanced cost center integration.
  - Automatically assigns cost center details to expense transactions when the cost center option is enabled.
  - Updated the method for retrieving enabled expense fields in the expense splitting functionality.

### DIFF
--- a/src/app/fyle/split-expense/split-expense.page.ts
+++ b/src/app/fyle/split-expense/split-expense.page.ts
@@ -52,7 +52,6 @@ import { SplitConfig } from 'src/app/core/models/split-config.model';
 import { ReviewSplitExpenseComponent } from 'src/app/shared/components/review-split-expense/review-split-expense.component';
 import { FyMsgPopoverComponent } from 'src/app/shared/components/fy-msg-popover/fy-msg-popover.component';
 import { SplittingExpenseProperties } from 'src/app/core/models/splitting-expense-properties.model';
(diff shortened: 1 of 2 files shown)


src/app/fyle/split-expense/split-expense.page.ts
```diff
@@ -1336,6 +1339,18 @@ export class SplitExpensePage implements OnDestroy {
     }
   }
 
+  private addCostCenterIdToTxnFields(): void {
+    if (this.txnFields.cost_center_id) {
+      return;
+    }
+    this.expenseFieldsService.getAllEnabled().subscribe((res: ExpenseField[]) => {
```
Dimple16, line 1346: Why make another API call here? Can we not pass the expense fields from add-edit-expense page?

Dimple16 (changes requested):
important comment

src/app/fyle/split-expense/split-expense.page.ts
Sishhhh: The expense fields that were coming from the add-edit-expense page is actually having the data of custom field, they are making the call of` is_custom: 'eq.true'`

Dimple16 (approved):

Chethan-Fyle (approved):

PR #3605, opened 2025-03-17
fix: walkthrough overlay click logic

amit429:
## Clickup
https://app.clickup.com/t/86cy9x766

## Code Coverage
Please add code coverage here

## UI Preview

## Summary by CodeRabbit

- **New Features**
  - Enhanced the dashboard guided tour experience by refining overlay interactions. The tour behavior now adjusts based on overlay clicks, allowing for smoother navigation and improved tracking of user interactions.
  - Introduced new properties and methods to manage the state of overlay interactions, enhancing the functionality of the dashboard.

- **Bug Fixes**
  - Updated method signatures to ensure correct handling of feature configuration related to the navbar walkthrough.


src/app/fyle/dashboard/dashboard.page.ts
```diff
@@ -155,23 +181,40 @@ export class DashboardPage {
       doneBtnText: 'Ok',
       nextBtnText: 'Next',
       prevBtnText: 'Back',
+      onCloseClick: () => {
```
arjunaj5, line 184: this means the overlay outside the walkthrough is clicked is it?

src/app/fyle/dashboard/dashboard.page.ts
```diff
@@ -155,23 +181,40 @@ export class DashboardPage {
...
+        this.walkthroughService.setIsOverlayClicked(false);
+        this.setNavbarWalkthroughFeatureConfigFlag(false);
+        driverInstance.destroy();
+      },
+      onDeselected: () => {
```
arjunaj5, line 189: what does this mean for walkthrough?

src/app/fyle/dashboard/dashboard.page.ts
```diff
@@ -155,23 +181,40 @@ export class DashboardPage {
...
+        const activeIndex = driverInstance.getActiveIndex();
+        if (activeIndex) {
+          this.walkthroughService.setActiveWalkthroughIndex(activeIndex);
+        }
+      },
+      onNextClick: () => {
+        driverInstance.moveNext();
+        if (this.walkthroughService.getActiveWalkthroughIndex() === navbarWalkthroughSteps.length - 1) {
+          this.walkthroughService.setIsOverlayClicked(false);
```
arjunaj5, line 198: can you add a code comment why this is needed?

src/app/fyle/dashboard/dashboard.page.ts
```diff
@@ -182,10 +225,17 @@ export class DashboardPage {
...
+        isShown?: boolean;
+        isFinished?: boolean;
+        overlayClickCount?: number;
+        currentStepIndex?: number;
+      }>(showNavbarWalkthroughConfig)
       .subscribe((config) => {
         const featureConfigValue = config?.value || {};
         const isFinished = featureConfigValue?.isFinished || false;
+        this.overlayClickCount = featureConfigValue?.overlayClickCount || 0;
+        this.walkthroughOverlayStartIndex = featureConfigValue?.currentStepIndex || 0;
```
arjunaj5, line 238: this is actually `walkthroughOverlayCurrentStepIndex` no?

arjunaj5 (commented):
Added comments and questions.

src/app/fyle/dashboard/dashboard.page.ts
amit429: onCloseClick this is a callback function of driver.js when the cross arrow for the walkthrough is clicked

src/app/fyle/dashboard/dashboard.page.ts
amit429: Yes the walkthroughOverlayStartIndex is same as walkthroughOverlayCurrentStepIndex but according to the condition , we want to start the walkthrough from where the user left when the user clicks on the overlay for the first time , thats why I have named it walkthroughOverlayStartIndex

src/app/fyle/dashboard/dashboard.page.ts
amit429: onNextClick this is needed now because we are also monitoring the overlay click , we need to check that if the last button in the walkthrough is getting clicked and the walkthrough is getting destroyed , then we should handle the feature config save accordingly. Will add a comment in the code where ever necessary

src/app/fyle/dashboard/dashboard.page.ts
amit429: onDeselected this is a callback function of driver.js which gets triggered for every deselected step and move for the next step. I want it to register every index that go completed by walkthrough

src/app/fyle/dashboard/dashboard.page.ts
```diff
@@ -82,6 +82,12 @@ export class DashboardPage {
 
   isWalkthroughPaused = false;
 
+  isOverlayClicked = true;
```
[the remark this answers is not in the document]
amit429: Its needed as there is no way to find out the overlay click behaviour for driver.js

src/app/core/services/walkthrough.service.ts
```diff
@@ -70,4 +72,12 @@ export class WalkthroughService {
...
     this.activeWalkthroughIndex = index;
   }
+
+  setIsOverlayClicked(isClicked: boolean): void {
+    this.isOverlayClicked = isClicked;
+  }
+
+  getIsOverlayClicked(): boolean {
+    return this.isOverlayClicked;
+  }
```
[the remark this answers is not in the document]
amit429: The overlay clicked for any user is particular to that session only , in the middle of walkthrough if the user closes the app , the walkthrough will start again and for the overlay click we are maintaining a feature config variable if during the session the user clicks on the walkthrough

src/app/core/services/walkthrough.service.ts
```diff
@@ -7,6 +7,8 @@ import { DriveStep } from 'driver.js';
 export class WalkthroughService {
   activeWalkthroughIndex = 0;
 
+  isOverlayClicked = true;
```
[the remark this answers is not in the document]
amit429: Here so now the overlayclicked destroys the walkthrough, its true by default because the logic is when the user enters there is no way to identify if the user clicked on the overlay or not for driver js, so the logic is implemented such that if the walkthrough is destroyed by user when they finish or close then the overlayclicked becomes false , or if the walkthrough completes without the finish button or close button click , then we will consider it as overlayclick finish , thats why its made true by default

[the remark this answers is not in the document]
amit429: isOverlayclicked is a check for weather the user clicked on the overlay bg outside the walkthrough pop up

arjunaj5 (approved):

PR #3607, opened 2025-03-18
fix: show correct msg for disabled category

Sishhhh:
## Clickup
https://app.clickup.com/t/86cxubfep

## Summary by CodeRabbit

- **New Features**
  - The expense splitting screen now displays clearer, more helpful messages when category or cost center options aren’t available, guiding users on what to do next.
  - Introduced a new message property to provide feedback when no category is assigned.

### DIFF
--- a/src/app/fyle/split-expense/split-expense.page.ts
+++ b/src/app/fyle/split-expense/split-expense.page.ts
@@ -144,6 +144,8 @@ export class SplitExpensePage implements OnDestroy {
 
   isReviewModalOpen = false;
 
+  disableMessage = '';
+
   private splitExpenseData: Subscription;
 
   constructor(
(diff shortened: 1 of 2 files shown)


src/app/fyle/split-expense/split-expense.page.ts
```diff
@@ -144,6 +144,8 @@ export class SplitExpensePage implements OnDestroy {
 
   isReviewModalOpen = false;
 
+  disableMessage = '';
```
arjunaj5, line 147: does not make much sense in the split expense page, can you come up with a better name?

src/app/fyle/split-expense/split-expense.page.ts
```diff
@@ -1028,14 +1030,17 @@ export class SplitExpensePage implements OnDestroy {
...
-        const control = this.splitExpensesFormArray.at(i);
-        const categoryControl = control?.get('category');
-        if (categoryControl) {
-          categoryControl.clearValidators();
-          categoryControl.updateValueAndValidity();
+    if (categories.length === 0) {
+      this.disableMessage = 'No category is assigned. Please contact admin for further help.';
+      if (this.splitConfig.category.is_mandatory) {
+        this.splitConfig.category.is_mandatory = false;
+        for (let i = 0; i < 2; i++) {
```
arjunaj5, line 1037: will this work for more than 2 splits?

arjunaj5 (changes requested):

src/app/fyle/split-expense/split-expense.page.ts
Sishhhh: When the UI renders, it initially displays two splits by default. Their required validation is determined by `splitConfig.category.is_mandatory`. By setting it to `false`, any new splits added later will not have the mandatory validator, as we add a new split using:
```
this.formBuilder.control(
  isFirstSplit ? this.splitConfig.category.value : '',
  this.splitConfig.category.is_mandatory ? [Validators.required] : null
)
```

Thus, we are only removing the required validator for the first two splits, as they were initialized with the old splitConfig value

arjunaj5 (approved):

PR #3667, opened 2025-04-24
fix: Updated the creator_id in expense to user_id

devendrafyle:
## Clickup
https://app.clickup.com/

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved accuracy of expense creator identification in expense-related features.


src/app/core/services/transaction.service.ts
```diff
@@ -688,7 +688,7 @@ export class TransactionService {
         flight_return_travel_class: expense.travel_classes && expense.travel_classes[1],
         hotel_is_breakfast_provided: expense.hotel_is_breakfast_provided,
         tax_group_id: expense.tax_group_id,
-        creator_id: expense.employee?.id,
+        creator_id: expense.creator_user_id,
```
devendrafyle, line 691: this key has only one use case in verify the latest User comment on policy violated expense, so updated this as expense comments have `user_id` in them instead of `org_user_id` or `employee_id`.

arjunaj5 (approved):

PR #3674, opened 2025-05-06
feat: Migrated the transaction delete api to platform

devendrafyle:
## Clickup
https://app.clickup.com/t/86cyvk0eq

## Summary by CodeRabbit

- **New Features**
  - Added support for deleting multiple expenses at once via a new bulk deletion method.

- **Bug Fixes**
  - Improved consistency in expense deletion across different sections of the app.

- **Refactor**
  - Unified expense deletion logic by replacing multiple deletion methods with a single bulk deletion method.
  - Updated deletion flows in expense, mileage, and per diem pages to use the new bulk deletion approach.
  - Adjusted related tests and service mocks to align with the updated deletion method.


src/app/core/services/platform/v1/spender/expenses.service.ts
```diff
@@ -376,4 +376,14 @@ export class ExpensesService {
...
   }
+
+  deleteExpenses(expenseIds: string[]): Observable<void> {
+    const payload = {
+      data: expenseIds.map((id) => ({ id })),
+    };
+
+    return this.spenderService
+      .post<PlatformApiResponse<null>>('/expenses/delete/bulk', payload)
+      .pipe(map((): void => void 0));
```
arjunaj5, line 387: can we change the return type to Observable<null> and replace this with `map(response => response.data)` ?

arjunaj5 (approved):
added one question. Rest LGTM

src/app/core/services/platform/v1/spender/expenses.service.ts
devendrafyle: this API doesn't return any response, thats why, it only return HTTP status of 200 on success.

Chethan-Fyle:
```
return this.spenderService
  .post<unknown>('/expenses/delete/bulk', payload)
  .pipe(map(() => {}));
```

Chethan-Fyle (approved):

PR #3725, opened 2025-06-18
fix: migrate save api for advances to platform

RishabhSinha02:
## Clickup
[Please add link here](https://app.clickup.com/t/86czap0uy)

## Code Coverage
Please add code coverage here

## UI Preview


src/app/core/services/advance-request.service.ts
```diff
@@ -76,7 +77,8 @@ export class AdvanceRequestService {
     private dateService: DateService,
     private fileService: FileService,
     private approverService: ApproverService,
-    private spenderService: SpenderService
+    private spenderService: SpenderService,
+    private spenderPlatformV1ApiService: SpenderPlatformV1ApiService
```
mvaishnavi, line 81: let's use `spenderService` - I don't think we need the other service injection here.

PR #3727, opened 2025-06-19
fix: migrate action permissions advance request api to platform

RishabhSinha02:
## Clickup
[Please add link here](https://app.clickup.com/t/86czaqgdq)

## Code Coverage
Please add code coverage here

## UI Preview

### DIFF
--- a/src/app/core/models/popover-data.model.ts
+++ b/src/app/core/models/popover-data.model.ts
@@ -1,0 +1,5 @@
+export interface PopoverData {
+  comment?: string;
+  status?: string;
+  updated?: boolean;
+}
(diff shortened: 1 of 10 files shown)


src/app/core/models/popover-data.model.ts
```diff
@@ -0,0 +1,5 @@
+export interface PopoverData {
+  comment?: string;
+  status?: string;
+  updated?: boolean;
+}
```
arjunaj5, line 5: PopoverData name is too generic, is this specific to advances  API? please name it with advances in name.

are all fields optional?

arjunaj5 (approved):
minor comment and question

src/app/core/models/popover-data.model.ts
RishabhSinha02: each field is only relevant to a specific case, which is why they are all marked optional.
`comment` is only used in the pullback popover
`status` is used in the delete popover
`updated` is used in the comments modal

and i renamed PopoverData to AdvanceRequestPopoverData

PR #3742, opened 2025-06-25
feat:aggregate /reports/stats using group_by params to reduce 429s

aniket607:
## Clickup
[Please add link here
](https://app.clickup.com/t/86czcuv96)


src/app/fyle/dashboard/dashboard.service.ts
```diff
@@ -111,4 +101,16 @@ export class DashboardService {
         return 'Processing';
     }
   }
+
+  private transformStat(stat?: GroupedReportStatsResponse): PlatformReportsStatsResponse {
```
devendrafyle, line 105: GroupedReportStatsResponse => GroupedReportStats

PR #3776, opened 2025-07-16
fixes for mat form field

arjunaj5:
## Clickup
https://app.clickup.com

## Code Coverage
Please add code coverage here

## UI Preview


src/app/shared/components/send-email/send-email.component.html
```diff
@@ -1,64 +0,0 @@
-<ion-content>
```
arjunaj5, line 1: unused component

PR #3924, opened 2025-09-05
feat: migrate /vendors/all -> /role/merchants

SahilK-027:
## Clickup
app.clickup.com

### DIFF
--- a/src/app/core/models/platform/platform-merchants.model.ts
+++ b/src/app/core/models/platform/platform-merchants.model.ts
@@ -1,0 +1,7 @@
+export interface PlatformMerchant {
+  id: number;
+  created_at: Date;
+  updated_at: Date;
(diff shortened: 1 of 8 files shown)


src/app/core/models/platform/platform-merchants.model.ts
```diff
@@ -0,0 +1,7 @@
+export interface PlatformMerchant {
```
Dimple16, line 1: Why name it Platform Merchants? We'll eventually move to a world where everything is platform. Do we have any other existing merchant model?

Dimple16: We can name it `Merchant` . Please update the file name accordingly.

Dimple16 (approved):
Minor comments. Please fix before merging

